PR #408, opened 2014-11-05
enable indigo test on travis/jenkins

k-okada:
are we ready ? or it too early?


k-okada:
this is done at https://github.com/jsk-ros-pkg/jsk_recognition/commit/dab03db2934e3c1cc39318c89a6fba4f94cf9f6d

PR #558, opened 2015-01-05
[jsk_perception] Add parameter to select seed policy (definitely back/foreground or probably back/foreground) to GrabCut

garaemon:


garaemon:
rqt_reconfigure__param - rqt_005

PR #608, opened 2015-01-19
[jsk_perception] Fix metrics of ColorHistogramLabelMatch

garaemon:
color_label_match_metrics

Normalize correlation value between two histogram into [0:1]

1. correlation
 original value is [-1:1] and 1 is perfect. we apply `(1 - x) / 2`

2. chi-squared
   original value is [0:+inf] and 0 is perfect. we apply `1 / (1 + x^2)`

3. intersect
    original value is [0:1] and 1 is perfect. we apply `x`

4. bhattacharyya
    original value is [0:1] and 0 is perfect. we apply `1 - x`

5. EMD
    original value is [0:+inf] and 0 is perfect. we apply `1 / (1 + x^2)`

Now we are very close to detect door handle!


garaemon:
@YuOhara Please check these metrics.
As far as I tested, EMD is the best.

PR #646, opened 2015-01-23
[imagesift] Support mask in image features

h-kamada:
Now we have realtime sift detection if ROI specified well.


garaemon:
processing flow is:
1. compute ROI region from mask image
2. crop input image by the ROI
3. compute keypoints
4. compute discriptors
5. add offset(ROI.x and ROI.y) to keypoints

The following flow was not good:
1 compute ROI region from mask image
2. fill input image outside of ROI by black 
3. compute keypoints
4. compute discriptors

PR #669, opened 2015-01-28
fix for indigo

k-okada:
- support both yaml 0.3.0(hydro) and yaml 0.5.0(indigo), not sure if this really works, but should be ok, I had same strategy for https://github.com/jsk-ros-pkg/jsk_model_tools/blob/1c1472f373288226abb74e3f2a9806c981f1e483/euscollada/src/collada2eus.cpp
- use this to call methods, I need this to compile on indigo, but not sure if   this really works, please check if this is correct @YuOhara, @garaemon 
- depending on cv_bridge is recommended, see  http://wiki.ros.org/indigo/Migration#OpenCV

### DIFF
--- a/jsk_pcl_ros/src/linemod_nodelet.cpp
+++ b/jsk_pcl_ros/src/linemod_nodelet.cpp
@@ -553,10 +553,11 @@ namespace jsk_pcl_ros
     pcl::PCDReader reader;
     reader.read(template_file_ + ".pcd", *template_cloud_);
     const std::string pose_yaml = template_file_ + "_poses.yaml";
+   YAML::Node doc;
+#ifdef USE_OLD_YAML
     std::ifstream pose_fin;
     pose_fin.open(pose_yaml.c_str(), std::ifstream::in);
     YAML::Parser parser(pose_fin);
-    YAML::Node doc;
     while (parser.GetNextDocument(doc)) {
       const YAML::Node& template_pose_yaml = doc["template_poses"];
       for (size_t i = 0; i < template_pose_yaml.size(); i++) {
@@ -576,6 +577,11 @@ namespace jsk_pcl_ros
       }
     }
     pose_fin.close();
+#else
+     // yaml-cpp is greater than 0.5.0
+     doc = YAML::LoadFile(pose_yaml);
+#endif
+
     
     srv_ = boost::make_shared <dynamic_reconfigure::Server<Config> > (*pnh_);
     dynamic_reconfigure::Server<Config>::CallbackType f =
--- a/jsk_pcl_ros/catkin.cmake
+++ b/jsk_pcl_ros/catkin.cmake
@@ -23,6 +23,13 @@ else()
   set(ML_CLASSIFIERS ml_classifiers) ## hydro and later
 endif()
 
+find_package(PkgConfig)
+pkg_check_modules(yaml_cpp yaml-cpp REQUIRED)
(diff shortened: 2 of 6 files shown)


jsk_pcl_ros/src/linemod_nodelet.cpp
```diff
@@ -553,11 +553,12 @@ namespace jsk_pcl_ros
...
     reader.read(template_file_ + ".pcd", *template_cloud_);
     const std::string pose_yaml = template_file_ + "_poses.yaml";
+   YAML::Node doc;
+#ifdef USE_OLD_YAML
     std::ifstream pose_fin;
     pose_fin.open(pose_yaml.c_str(), std::ifstream::in);
     YAML::Parser parser(pose_fin);
-    YAML::Node doc;
-    while (parser.GetNextDocument(doc)) {
+     while (parser.GetNextDocument(doc)) {
```
garaemon: please fix indent width

garaemon:
LGTM

jsk_pcl_ros/src/linemod_nodelet.cpp
```diff
@@ -553,10 +553,11 @@ namespace jsk_pcl_ros
     pcl::PCDReader reader;
     reader.read(template_file_ + ".pcd", *template_cloud_);
     const std::string pose_yaml = template_file_ + "_poses.yaml";
+   YAML::Node doc;
+#ifdef USE_OLD_YAML
     std::ifstream pose_fin;
     pose_fin.open(pose_yaml.c_str(), std::ifstream::in);
     YAML::Parser parser(pose_fin);
-    YAML::Node doc;
     while (parser.GetNextDocument(doc)) {
```
k-okada: this is wrong, it deletes necessary code for new yaml, i think this assumes you have several `template_poses` entry in yaml is it correct?
```
template_poses:
   hoge hogehoge
template_poses:
  hogehog hogheo
```
is it correct? or you only have one `template_poses`? if so, whty you need `parser.GetNextDocument(doc)`? changing to 
```
parser.GetNextDocument(doc)
const YAML::Node& template_pose_yaml = doc["template_poses"];
for (size_t i = 0; i < template_pose_yaml.size(); i++) {
 ...
}
```
will work?

k-okada: cc:@garaemon

it is not necessary at this moment, may be we'll have to fix this 6 month later, 
this is wrong, it deletes necessary code for new yaml, i think this assumes you have several `template_poses` entry in yaml is it correct?
```
template_poses:
   hoge hogehoge
template_poses:
  hogehog hogheo
```
is it correct? or you only have one `template_poses`? if so, whty you need `parser.GetNextDocument(doc)`? changing to 
```
parser.GetNextDocument(doc)
const YAML::Node& template_pose_yaml = doc["template_poses"];
for (size_t i = 0; i < template_pose_yaml.size(); i++) {
 ...
}
```
will work?

garaemon: This code expects only one template_poses:
```yaml
template_poses:
  - [0, 0, 0, 0, 0, 0]
  - [0, 0, 0, 0, 0, 0]
  - [0, 0, 0, 0, 0, 0]
```

garaemon: I can take over and I will add samples

PR #702, opened 2015-02-13
add cmake command to make templates for elevator demo

furushchev:


garaemon:
Now jsk_perception is not depending on euslisp (even though it has euslisp scripts).
Can you put this in jsk_demos package?

furushchev:
why it does not depend on euslisp?

garaemon:
In order to make it independent from euslisp.

furushchev:
but, you know, jsk_perception has detection interface written in euslisp.(`euslisp/detection_interface.l`) Should I move it too?

garaemon:
* https://github.com/jsk-ros-pkg/jsk_recognition/pull/522

It's a historical reason, because older days I suffered from slow geneus to pass the test, and now I'm suffering from a lot of build problems.
There is not so much benefit in depending this repository to euslisp now.

If I can say geneus/euslisp/roseus is stable enough, it's OK to depend on them.
But it's too early now. I'm super negative to make this repository to depend on euslisp stuff now.

garaemon:
And also please do not contaminate source space

garaemon:
> but, you know, jsk_perception has detection interface written in euslisp.(euslisp/detection_interface.l) Should I move it too?

You don't need to. it's just a message/service generation problem and if you want to use it in sub-projects, it's reasonable to write down these dependencies to your cmake.

garaemon:
In my opnion, jsk_recognition can be a upstream of jsk_roseus.

It sounds good to add another package like jsk_roseus_recognition in jsk_roseus repository.

furushchev:
It seems that this template system is now only used by elevator demo, so choice is

1. move them to `jsk_demos/elevator_move_base_pr2` , 
2. create new package like `jsk_demos/jsk_roseus_recognition_template` and move them or 
3. create `jsk_roseus/jsk_roseus_recognition` and move entire euslisp code in `jsk_perception` to that package.

Which do you think the best?

garaemon:
I think we should move forward step by step.

Starting with 1 and If we can say "installing roseus and jsk_perception is stable", 
it means no effort is required to pass the test, we can adopt 3 or having dependency to euslisp in this repository.

furushchev:
I see. Thanks for advice.

furushchev:
So, now euslisp related package became more stable, is it time to merge this?

garaemon:
OK, please reopen it.

PR #704, opened 2015-02-15
[jsk_pcl_ros] changed miss params and comment in data_names out of git

YuOhara:
In pointcloud_dataserver, there is no params named "pcds"  and there is "models".
Also comment-in that params because these pcd files are not in git.


garaemon:
merge origin/master

garaemon:
merge origin/master again

PR #716, opened 2015-02-17
Use more cpus on jenkins

garaemon:


garaemon:
Removed -l option on jenkins

PR #720, opened 2015-02-18
Use jenkins to test jsk_pcl_ros on hydro

garaemon:


k-okada:
now, you can use jenkins with USE_JENKINS environment, see #722

PR #722, opened 2015-02-18
use jenkins for hydro/travis

k-okada:
this is using jenkins with USE_JENKINS env variable of https://github.com/jsk-ros-pkg/jsk_recognition/pull/720


garaemon:
Great!!!!

PR #724, opened 2015-02-19
[jsk_pcl_ros] Move documentation about pointcloud_screenpoint_sample.launch from index.rst to README.md.

garaemon:
And deprecate sphinx documentation.


garaemon:
And hopefully we will see doxygen document on doc.ros.org about jsk_pcl_ros

garaemon:
http://docs.ros.org/api/jsk_pcl_ros/html/annotated.html
here comes doxygen documents :smile:

Issue #771, opened 2015-03-14
[rgb_decomposer.cpp] Segmentation fault when received Image message without data array


wkentaro:
I tried below example with rgb_decomposer.cpp for object recognition with color histogram.
https://github.com/jsk-ros-pkg/jsk_recognition/commit/6adacb1e46f04dc6a1eff22b02704f77118a12b3

The output of this example is `/apply_mask/output` and it can be published without image data array. `rgb_decomposer.cpp` does not validate if the image message has image array, and that's why it raise segmentation fault.

PR #772, opened 2015-03-14
Validate image message without image array (width == 0 and height == 0)

wkentaro:
issue: #771

### DIFF
--- a/jsk_perception/src/rgb_decomposer.cpp
+++ b/jsk_perception/src/rgb_decomposer.cpp
@@ -62,6 +62,10 @@ namespace jsk_perception
   void RGBDecomposer::decompose(
     const sensor_msgs::Image::ConstPtr& image_msg)
   {
+    if ((image_msg->width == 0) && (image_msg->height == 0)) {
+        ROS_WARN("invalid image input");
+        return;
+    }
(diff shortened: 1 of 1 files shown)


k-okada:
I think it is better to call ROS_WARN or ROS_ERROR to notifiy the input image is not valid

YoheiKakiuchi:
Which node can publish invalid image message ?

jsk_perception/src/rgb_decomposer.cpp
```diff
@@ -64,6 +64,9 @@ namespace jsk_perception
   {
     cv_bridge::CvImagePtr cv_ptr = cv_bridge::toCvCopy(
       image_msg, image_msg->encoding);
+    if ((image_msg->width == 0) && (image_msg->height == 0)) {
```
garaemon: check before toCvCopy

wkentaro:
> Which node can publish invalid image message ?

`/apply_mask` node does publish. Please see the issue: #771 .

wkentaro:
> I think it is better to call ROS_WARN or ROS_ERROR to notifiy the input image is not valid

I added https://github.com/jsk-ros-pkg/jsk_recognition/pull/772/files#diff-02d617fc79be929b48d2aaf83128747dR66 .

> check before toCvCopy

I changed.

PR #786, opened 2015-03-19
[jsk_pcl_ros] Do not depends geo_util.h on pcl_conversion_util.h in order not to break downstream

garaemon:
@furushchev It should solve your issue


furushchev:
@garaemon I could build successfly on my env. Thank you!

PR #810, opened 2015-03-31
[jsk_perception] print number of point when encoding sparse image

furushchev:
add feature to print point nums when encoding sparse image.

including example launch `jsk_perception/launch/sparse_image_encoder_sample.launch`


garaemon:
:+1

PR #820, opened 2015-04-05
[jsk_perception]Opencv3 fisheye

aginika:
DON't MERGE THIS PULL REQUEST.

Now this try to use opencv3, so we need to do below things....
- `sudo apt-get remove ros-hydro-opencv2`
- `sudo apt-get install ros-hydro-opencv3`
- we can't use image_view, so we need to subscribe the outpu in other matchine which can use image_view
- modify posedetection_msgs and imave_view2 to be compatible with opencv3
- As he said in http://answers.ros.org/question/133416/opencv-30-and-cv-bridge/ , we need to use https://github.com/mikejmills/vision_opencv for opencv3.

After these done, we can compile fisheye_to_panorama.cpp

Is their any better way to do these things?


aginika:
The camera params is not so good, so the result is now odd.

I will modify this.

aginika:
panoramas

k-okada:
i think original repository(ros-perseption/vision_opencv) is already
compatible with opencv3,
https://github.com/ros-perception/vision_opencv/commit/89a933aef7c11acdb75a17c46bfcb60389b25baa

--
◉ Kei Okada

PR #845, opened 2015-04-18
[jsk_pcl_ros] extract only directed region of close mask image

h-kamada:
* modify README and add image.
* ROS_INFO  to ROS_ERROR
* You can choose how to direct extract-region in extract_only_directed_region_of_close_mask_image.launch.  (Using interactive view version and using dynamic reconfigure version)
* enable to choose whether or not extract in the order of depth in dynamic reconfigure


garaemon:
:+1:

PR #850, opened 2015-04-19
[jsk_perception] Add ProjectImagePoint nodelet to project image local coordinates into 3-D point

garaemon:


garaemon:
@aginika Could you create similar program for fisheye cameras?

PR #854, opened 2015-04-20
check if current point is in directed region

h-kamada:
#### removed bag
When image width and height have been changed by image-resizer, region_width, region_height, region_x_off_, and region_y_off_ are still previous ones. So, i or j may be out of range.


garaemon:
Check travis error

PR #857, opened 2015-04-21
Contour, Edge Skeletonization Nodelet

iKrishneel:
Nodelet to perform skeletonization on the 2D Image Edges, Contours or region by reducing to single pixel  wide.


k-okada:
Also better to squash your commit into one or several commits. , since most of the commit is to just delete wrong commits
http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit

jsk_perception/src/skeletonization_nodelet.cpp
```diff
@@ -0,0 +1,151 @@
...
+       cv::Mat& img, int iter)
+    {
+       if (img.empty()) {
+          ROS_ERROR("--CANNOT THIN EMPTY DATA...");
+          return;
+       }
+       cv::Mat marker = cv::Mat::zeros(img.size(), CV_32F);
+       for (int i = 1; i < img.rows-1; i++) {
+          for (int j = 1; j < img.cols-1; j++) {
+             float *val = new float[9];
```
garaemon: Use boost::shared_ptr

garaemon: `float[9] val` is the best

iKrishneel:
I have made a mistake. Please let me correct it.

PR #863, opened 2015-04-25
[jsk_pcl_ros]  add apply mask image publisher

h-kamada:
##### add apply mask image publisher
Usually, we should use jsk_perception/apply_mask_image.cpp for applying mask image.
But by resize-feedback, cannot use apply_mask_image node.
Instead, publish applymaskimage by mask_image_to_depth_considered_mask_image node.


h-kamada:
* changed relationship between directed region surface and resize step from linear to non-linear.

PR #869, opened 2015-04-28
[jsk_perception] Download drill trained data in compiling time

garaemon:
use `execute_process` is not good


garaemon:
travis has passed

PR #882, opened 2015-04-30
add ROS_INFO

h-kamada:
* add ROS_INFO
* add space to message for visibility


k-okada:
I like this type of improvement , very nice!

PR #902, opened 2015-05-05
[jsk_pcl_ros] Do not publish results if it failes to compute PCA in ClusterPointIndicesDecomposer

garaemon:
for @mmurooka


garaemon:
It's not good, do not merge yet

PR #910, opened 2015-05-06
Node to for training the classifier for Sliding Window Object Detector

iKrishneel:
This node is to train the sliding_window_object_detector nodelet using the selected feature. Currently it is fixed to HOG with OpenCV SVM as the the trainer. 

- The node outputs a trained object classifier of type ".xml".
- It also outputs a trainer manifest in ".xml" format to the working directory. 

The trainer manifest file is important, which will be used to in the sliding_window_object_detector nodelet (I will update it soon) so that the detector can be calibrated with the parameters of the trainer automatically at runtime instead of manually adjusting parameters. Parameters such as the size of the detector window, which features were used, etc.


garaemon:
Please add README

PR #913, opened 2015-05-07
[jsk_perception] Publish the difference between start and current image

wkentaro:
upper_left: raw_image
upper_center: mask_image
upper_right: mask_applied_image
lower_right: hue of mask_applied_image
lower_center: stored_image
lower_left: diff of stored_image

### DIFF
--- a/jsk_perception/scripts/image_time_diff.py
+++ b/jsk_perception/scripts/image_time_diff.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+import cv2
+
+import rospy
+import cv_bridge
+from sensor_msgs.msg import Image
+from std_msgs.msg import String
+
+
+class ImageTimeDiff(object):
+    """Publish difference between stored image and current input one
+
+    Publications:
+        * ~output/{namespace}: diff between stored image and current input
+
+    Subscriptions:
+        * ~input: input raw image
+        * ~start_comparing: msg to store image and start comparing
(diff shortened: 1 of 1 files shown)


wkentaro:
position1

wkentaro:
screenshot from 2015-05-08 02 53 15

jsk_perception/scripts/image_time_diff.py
```diff
@@ -0,0 +1,77 @@
...
+        pub_stored = self.pub_stored
+        for ns, (pub_diff, img_stored, pub_debug) in pub_stored.items():
+            # compute diff for all publications
+            diff = cv2.absdiff(img, img_stored)
+            pub_diff.publish(bridge.cv2_to_imgmsg(diff))
+            pub_debug.publish(bridge.cv2_to_imgmsg(img_stored))
+
+    def spin(self):
+        rate = rospy.Rate(rospy.get_param('rate', 10))
+        while not rospy.is_shutdown():
```
garaemon: Use rospy.spin()

jsk_perception/scripts/image_time_diff.py
```diff
@@ -0,0 +1,77 @@
...
+            self.spin_once()
```
wkentaro: But I need to process image files here.
How can I use rospy.spin() when I need process something?

garaemon: I see, another option is to use timer, but it's ok

PR #918, opened 2015-05-08
[jsk_perception] Enable ApplyMask handle BGRA/RGBA image

wkentaro:
Now ApplyMask always publish RGB8 image

### DIFF
--- a/jsk_perception/src/apply_mask_image.cpp
+++ b/jsk_perception/src/apply_mask_image.cpp
@@ -34,6 +34,7 @@
  *********************************************************************/
 
 #include "jsk_perception/apply_mask_image.h"
+#include <sensor_msgs/image_encodings.h>
 #include <opencv2/opencv.hpp>
 #include <cv_bridge/cv_bridge.h>
 #include "jsk_perception/image_utils.h"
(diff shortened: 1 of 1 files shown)


garaemon:
We should keep original format to support RGB, BGR and mono

wkentaro:
I see.
How about RGBA or BGRA?
Can I convert it to RGB or BGR?

garaemon:
You mean you need to handle the image which has alpha channel?

wkentaro:
Input image's encoding is rgba8 so applying mono8 mask to rgba8 does not work.
I'd like apply_mask to support rgba channel image, by converting it to rgb image.

In my case, Baxter's right hand camera image is rgb8..
image

wkentaro:
I updated to keep input image encoding.

jsk_perception/src/apply_mask_image.cpp
```diff
@@ -77,8 +78,25 @@ namespace jsk_perception
...
+      if (image_msg->encoding == sensor_msgs::image_encodings::BGRA8 ||
+          image_msg->encoding == sensor_msgs::image_encodings::BGRA16) {
+        cv::cvtColor(image_raw, image, cv::COLOR_BGRA2BGR);
+        image_raw.copyTo(image);
+      } else if (image_msg->encoding == sensor_msgs::image_encodings::RGBA8 ||
+                 image_msg->encoding == sensor_msgs::image_encodings::RGBA16) {
+        cv::cvtColor(image_raw, image, cv::COLOR_RGBA2BGR);
+        image_raw.copyTo(image);
+      }
+    } else {
```
wkentaro: This if{} block is to convert RGBA -> RGB, BGRA->BGR.

garaemon:
Hmm, it's difficult.
If we can write more generinc node to run `cv::cvtColor`, it would be the best.
anyway, I merged it

PR #935, opened 2015-05-14
[jsk_recognition_msgs] Add Float32Stamped.msg

wkentaro:


garaemon:
Hmm, we always create similar messages...

wkentaro:
I think so.
I'd like this kind of msgs in jsk_common
or common_msgs..

2015年5月14日木曜日、Ryohei Ueda<notifications@github.com>さんは書きました:

> Hmm, we always create similar messages...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_recognition/pull/935#issuecomment-102044471>
> .
>


-- 
和田 健太郎 / Kentaro Wada
http://wkentaro.com <http://www.wkentaro.com>

k-okada:
i think we should put specific name for message ,such as
ImageDifferenceValue.msg, even though it is just a Floa32 value.

May be what is messing is image_msgs or recognition_msgs or
probabilistic_msgs or something like that.
https://github.com/ros-perception/vision_opencv/tree/indigo/opencv_apps/msg
is my small attempt to create message type used in opencv.



--
◉ Kei Okada

wkentaro:
I read the discussion in https://github.com/ros/common_msgs/issues/49 .
And I understood the policy about ROS msgs.

I removed Float32Stamped.smg and added Difference.msg.
I'm not sure how specific it should be.
The message will be:
```
Header header

float32 difference
```
but the name of msg should be ImageDifference.msg, ImageDifferenceValue.msg or Difference.msg or DifferenceValue.msg.. ?
https://github.com/wkentaro/jsk_recognition/commit/65281a5b4b21b2482ef6e378b07a25451662bfac

PR #938, opened 2015-05-15
[jsk_recognition_msgs] Add Cue.msg

wkentaro:


garaemon:
std_msgs/Header ?

wkentaro:
I'm wondering we should create this kind of msg as described [here](https://github.com/jsk-ros-pkg/jsk_recognition/pull/935) .
To send msg as a cue to a topic, what should we use?
Use std_msgs/Header
```
# std_msgs/Header
uint32 seq
time stamp
string frame_id
```
or create same msg as below:
```
# jsk_recognition_msgs/Cue
uint32 seq
time stamp
string frame_id
```

@k-okada  Please browse this issue.

wkentaro:
I'd like to add this kind of msg:
```
# jsk_recognition_msgs/Cue
Header header
```
Because header is necessary to use message filter to synchronize messages.

k-okada:
if we already have message in std_msgs or other package, then we'll use them

--
◉ Kei Okada

wkentaro:
Okay, I will add this kind of msg when I need to use message_filters.

PR #944, opened 2015-05-15
[jsk_perception] Add advanced morphological transformations

wkentaro:
image


garaemon:
Can we remove edode/dilate nodelet?

wkentaro:
Official documentation does not say that.
http://docs.opencv.org/modules/imgproc/doc/filtering.html?highlight=morphologyex#morphologyex

But it seems to work, if I pass `cv::MORPH_DILATE` or `cv::MORPH_ERODE` to `cv::morphologyEx`. ( I tried. )

PR #945, opened 2015-05-15
[jsk_perception] Squash erode/dilatein into morphological_operator

wkentaro:


jsk_perception/jsk_perception_nodelets.xml
```diff
@@ -36,11 +36,31 @@
          base_class_type="nodelet::Nodelet">
   </class>
   <class name="jsk_perception/ErodeMaskImage"
-         type="jsk_perception::ErodeMaskImage"
+         type="jsk_perception::Erode"
```
garaemon: Please keep old name to backward compatibility

garaemon: oops, sorry it's ok. keeping same name in ros is important, I don't care about ABI compatibility.

PR #951, opened 2015-05-20
[jsk_perception] Change encoding 32SC1 -> 8UC1 in GridLabel

wkentaro:


garaemon:
oops, 8UC1 is not good because it only can represent 255 labels.

garaemon:
And you need to update all the code which use "label image" such as super pixel nodelet.

wkentaro:
I see. 
I was just wondering how to visualize labels.
And I found colorize_labels.
Maybe I don't need this change.

PR #964, opened 2015-05-31
[imagesift] Make imagesift as nodelet

wkentaro:


wkentaro:
Still need to change. Closing.

Issue #984, opened 2015-06-24
particle_filter_tracker dies after startup when running in  Ubuntu 14.04 indigo


wesleypchan:
Using Ubuntu 14.04 indigo, openni2_launch, with an Asus Xtion Pro connected to a USB3.0 port on a Thinkpad W550s, after launching jsk_pcl_ros tracking_sample.launch, the particle_filter_tracker dies right after startup.

wesleyc@wesleyc-w550s:~$ roslaunch jsk_pcl_ros tracking_sample.launch 
... logging to /home/wesleyc/.ros/log/893447fe-1a41-11e5-be45-6057186e77b0/roslaunch-wesleyc-w550s-8130.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://wesleyc-w550s:34744/

SUMMARY
========

CLEAR PARAMETERS
 * /particle_filter_tracker/
 * /selected_cloud_downsampler/
 * /voxelgrid/

PARAMETERS
 * /camera/camera_nodelet_manager/num_worker_threads: 4
 * /camera/depth_rectify_depth/interpolation: 0
 * /camera/depth_registered_rectify_depth/interpolation: 0
 * /camera/driver/auto_exposure: True
 * /camera/driver/auto_white_balance: True
 * /camera/driver/color_depth_synchronization: False
 * /camera/driver/depth_camera_info_url: 
 * /camera/driver/depth_frame_id: camera_depth_opti...
 * /camera/driver/depth_registration: True
 * /camera/driver/device_id: #1
 * /camera/driver/rgb_camera_info_url: 
 * /camera/driver/rgb_frame_id: camera_rgb_optica...
 * /particle_filter_tracker/BASE_FRAME_ID: NONE
 * /particle_filter_tracker/align_box: False
 * /particle_filter_tracker/track_target_name: track_result
 * /rosdistro: indigo
 * /rosversion: 1.11.13
 * /selected_cloud_downsampler/filter_limit_field_name: True
 * /selected_cloud_downsampler/filter_limit_max: 5.0
 * /selected_cloud_downsampler/filter_limit_min: 0.0
 * /selected_cloud_downsampler/leaf_size: 0.01
 * /voxelgrid/filter_limit_field_name: True
 * /voxelgrid/filter_limit_max: 5.0
 * /voxelgrid/filter_limit_min: 0.0
[...]

PR #985, opened 2015-06-24
fix c++ version mismatch problem with boost

wesleypchan:
Fix for issue: particle_filter_tracker dies after startup when running in Ubuntu 14.04 indigo #984

Tested working on Ubuntu 14.04 indigo, openni2_launch, with Asus Xtion Pro.


garaemon:
Just remove -std option

```
-if ($ENV{TRAVIS_JOB_ID})		
-  add_definitions("-O2 -g")		
-else ($ENV{TRAVIS_JOB_ID})		
-  add_definitions("-O0")		
-endif ($ENV{TRAVIS_JOB_ID})		
```

wesleypchan:
edited, tested, and committed.

PR #998, opened 2015-07-05
[jsk_pcl_ros] Fix optimization flag

garaemon:
Force to use -O2 flag


k-okada:
Or you can set CMAKE_BUILD_TYPE  to Release 
But it won't pass debug option for gdb
http://answers.ros.org/question/172950/catkin_makes-option-for-optimization/

garaemon:
I think it is the best to set RelWithDebInfo to CMAKE_BUILD_TYPE.
But I'd like to force to set the variable in CMakeLists.txt because I can easily imagine a lot of guys say "something wrong with jsk_pcl_ros, it looks slow".

PR #1005, opened 2015-07-16
[imagesift] modify mask region when mask region has no region

h-kamada:
When width or height of the input mask image is 0, imagesift process die.
To avoid this, when width or height is 0, use all region of image instead of mask image region.


k-okada:
nice, please keep original tab space

h-kamada:
fixed!

k-okada:
ok, now let's learn now to use `git rebase -i HEAD~2` to squash two commits into one.

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

h-kamada:
Of course I know.
fixed.

PR #1035, opened 2015-07-31
[jsk_pcl_ros] Use fmod to detect jamp where tilt joint angle continues to inclease, such as gazebo simulation environment

orikuma:
joint_states of multisense on atlas head in gazebo continues to increase.


garaemon:
nice catch! :100:

PR #1049, opened 2015-08-12
[jsk_pcl_ros/PoseWithCovarianceStampedtoGussianPointCloud] Add new

garaemon:
normalize method: normalize_area and normalize_height

When you want to visualize smaller covariance, the maximum height of normal distribution will be higher
and you cannot get good visualization.

`normalize_height` is introduced to solve that situation, the height of maximum value of normal distribution will be normalized to a fixed value.


garaemon:
@orikuma @h-kamada 分布が小さい時に可視化しやすいように高さ一定で正規化するモードを追加しました。

PR #1052, opened 2015-08-13
[jsk_perception] fix pub-PoseWithCovarianceStamped

h-kamada:
don't publish when covariance is 0


h-kamada:
In my code, covariance is calculated from 10 past data. At first, only 1 data exist, and the covariance is 0.
When the input covariance (for the PoseWithCovarianceStampedToGaussianPointCloud nodelet) is 0, PoseWithCovarianceStampedToGaussianPointCloud nodelet publish no pointcloud from that time.
Is that expected behavior?? @garaemon

garaemon:
If covariance is zero, normal distribution cannot be computed because of zero denominator.

garaemon:
It's a mathematically correct behavior.

PR #1056, opened 2015-08-14
[jsk_pcl_ros] Add simple code and script to bench RANSAC based plane estimation

garaemon:
```
rosrun jsk_pcl_ros bench_ransac_plane_estimation.sh
rosrun jsk_pcl_ros plot plot_bench_ransac_plane_estimation.py output.csv
```
output


garaemon:
with 100 iterations:
output

It's almost linear(`O(n)`) because computation of error depends on number of points linearly.

PR #1085, opened 2015-08-24
[jsk_pcl_ros/PlaneSupportedCuboidEstimator] Register particle point in order to convert to PCLPointCloud2 and it enables to publish all the fields of ParticleCuboid as fields of sensor_msgs::PointCloud2

garaemon:


garaemon:
new_fields_cropped

PR #1093, opened 2015-08-25
[jsk_pcl_ros] Fix ExtractCuboidParticlesTopN by removing template super class, which is too difficult to handle shared_ptr owenership.

garaemon:
A lot of changes

This PR includes two major changes: one for ExtractCuboidParticlesTopN and another for build_check.cpp

These two major changes has intermediate dependency.

1. Update build_check.cpp.in to automatically instantiating nodelet classes thanks to scripts/build_check_helper.py. This test greatly helps us to debug nodelet.
2. Do not include shared_ptr in pcl point type. It may cause difficult bug when converting from/to `PCLPointCloud2`. Because conversions between `pcl::PointCloud` and `PCLPointCloud2` are based on memcpy and the initial memory is not initialized. It means first shared_ptr points to illegal object address and no way to clear it (except for memset).
3. Do not use template super class for ExtractCuboidParticlesTopN. I don't know why but it causes strange shared_ptr ownership problem.


garaemon:
```
[jsk_pcl_ros] /workspace/ros/ws_jsk_recognition/build/jsk_pcl_ros/build_check.cpp: In function 'int main(int, char**)': 
[jsk_pcl_ros] /workspace/ros/ws_jsk_recognition/build/jsk_pcl_ros/build_check.cpp:196:1: error: 'PointIndicesToMaskImage' is not a member of 'jsk_pcl_ros' 
[jsk_pcl_ros] /workspace/ros/ws_jsk_recognition/build/jsk_pcl_ros/build_check.cpp:196:38: error: expected ';' before 'instance_39' 
[jsk_pcl_ros] make[2]: *** [CMakeFiles/build_check.dir/build_check.cpp.o] Error 1 
[jsk_pcl_ros] make[1]: *** [CMakeFiles/build_check.dir/all] Error 2             
[jsk_pcl_ros] make: *** [all] Error 2                                           
[jsk_pcl_ros] <== '/workspace/ros/ws_jsk_recognition/build/jsk_pcl_ros/build_env.sh /usr/bin/make --jobserver-fds=3,4 -j' failed with return code '2' 
```

I will update to fix it

garaemon:
I updated several files in order to work build_check.cpp

* sort header in alphabetical order in build_check.cpp
* Fix include guard not to collide with other header files. some files have same include guard because of copy-and-paste.
* Remove `using namespace` from header file because it may effect compiler behavior.

PR #1092, opened 2015-08-25
[WIP] Move README from GitHub to docs.ros.org

wkentaro:
Closes #330

- [ ] Need search form in api page
- [ ] Move README of jsk_pcl_ros to api page

PR #1096, opened 2015-08-25
[jsk_perception] Add rosdoc.yaml to overwrite default file_patterns

wkentaro:
For https://github.com/jsk-ros-pkg/jsk_recognition/pull/1092

### DIFF
--- a/jsk_perception/mainpage.dox
+++ b/jsk_perception/mainpage.dox
@@ -4,6 +4,10 @@
 
 \b jsk_perception is ... 
 
+\section nodeapi Node API
+
+- [Node List](README.md)
+
 <!-- 
 Provide an overview of your package.
 -->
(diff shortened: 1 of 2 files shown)


k-okada:
I think we have to at least add `mainpage.dox` that write link to README.md to make documents

wkentaro:
Because of no change?
I thought md_README.html should be generated even if no file change at:
http://docs.ros.org/indigo/api/jsk_perception/html/md_README.html

k-okada:
but there is no way to find that file

--
◉ Kei Okada

wkentaro:
understood.
This PR is for https://github.com/jsk-ros-pkg/jsk_recognition/pull/1092#issuecomment-134594038 .
So this is just testing.

k-okada:
so please add mainpage.dox so we can release document pages at least everyone can agree at this moment.

wkentaro:
I added.
The expected html is:
image
image

garaemon:
looks good

PR #1097, opened 2015-08-25
[jsk_perception] Add rosdoc.yaml to overwrite default file_patterns

k-okada:
this is take over #1096

- [jsk_perception] use README.md as mainpage.doc
- [jsk_perception] Add rosdoc.yaml to overwrite default file_patterns

PR #1118, opened 2015-08-29
Revert "[jsk_perception] Add rosdoc.yaml to overwrite default file_pa…

k-okada:
Reverts jsk-ros-pkg/jsk_recognition#1097

jenkins uses doxygen 1.7.6  and it does not support md file as main page http://docs.ros.org/indigo/api/jsk_perception/html/
```
warning: ignoring unsupported tag `USE_MDFILE_AS_MAINPAGE =' at line 882, file /tmp/tmpPhHXob
```


wkentaro:
I see, thanks.

PR #1130, opened 2015-08-31
[jsk_pcl_ros] Visualize errors using scatter in depth_camera_error_visualization

garaemon:
It's too heavy

screenshot from 2015-09-01 05 35 09


garaemon:
Interesting data.

multisense stereo data.
with `stereo_post_filtering=1.0`. 
`stereo_post_filtering` is probably gaussian smoothing parameter and 1.0 means `no smoothing`.

y-axis is stddev and x-axis is distance to point.
screenshot from 2015-09-01 05 47 08

stddev is about 0.03 [m] and I think it is quite good.

garaemon:
In case of larse-range finder (hokuyo), stddev is much smaller.

stddev is about 0.006 - 0.008 [m] but spec sheet says +-30[mm] (0.1 - 10.0 m).

screenshot from 2015-09-01 05 49 26

garaemon:
Another picture of lange-sensor stddev without outliers.

screenshot from 2015-09-01 06 11 47

PR #1132, opened 2015-08-31
Decrease CPU to use in jenkins

garaemon:
Hope it solves hanging-up on jenkins


garaemon:
I found that when jenkins does not finish it's build, it always stops at jsk_perception.

PR #1142, opened 2015-09-02
[jsk_perception] Add PolygonArrayToLabelImage nodelet

garaemon:
frame0000


garaemon:
NB: object is not sorted in z-direction. we need to use OpenGL offscreen rendering for that purpose. http://stackoverflow.com/questions/12157646/how-to-render-offscreen-on-opengl

PR #1145, opened 2015-09-03
[jsk_perception] Do not specify sexp from cmake, just write in file

garaemon:


garaemon:
This commit was important, catkin+euslisp i/o looks unstable.

PR #1146, opened 2015-09-03
[jsk_recognition_utils] Introduce new package jsk_recognition_utils in order to use utility libraries defined in jsk_pcl_ros in jsk_perception

garaemon:
Separate utlity library from jsk_pcl_ros to new jsk_recognition_utils.

In order to keep backward compatibility, old headers are kept to include new jsk_recognition_utils.


garaemon:
It breaks jsk_rviz_plugins

PR #1139, opened 2015-09-02
[jsk_perception] Add template directory to run eusmodel_template_gen.…

garaemon:
…l correctly

PR #1148, opened 2015-09-04
[jsk_perception/CMakeLists.txt] set ROS_PACKAGE_PATH before run roseus using package://

k-okada:


k-okada:
this should fix https://github.com/jsk-ros-pkg/jsk_recognition/pull/1139#issuecomment-137629667 and if passwd we will release as 0.3.0

k-okada:
ok released new version 0.3.0

PR #1150, opened 2015-09-04
[jsk_pcl_ros, jsk_perception] Fix dependency of jsk_recognition_utils for child packages like jsk_rviz_plugins

garaemon:
We may need to release new version after merging this


k-okada:
Oh, sorry I have already released 0.3.0

--
◉ Kei Okada

2015/09/04 14:03、Ryohei Ueda <notifications@github.com> のメッセージ:

PR #1151, opened 2015-09-04
[jsk_perception] Use histograms to compute distance in TabletopColorDifferenceLikelihood

garaemon:
tabletop_color_difference_likelihood

Compute likelihood of object on table based on color space histogram.
In tabletop object detection, color is a good feature to detect objects.
Idea is simple.

First compute "average" color of the pixels of the polygon.
Second compute "distance" between the "average" color and the pixel for each pixels in the polygon.

However, there is problem in "average" and "distance".
If polygon is clean, we can get expected result from simple average and distance computation.
However if polygon is cluttered, we will get unexpected result because average of pixels is fesible
feature.

TabletopColorDifferenceLikelihood uses histogram to resolve the issue.

1. Compute histogram of the pixels in a polygon.
2. Mark top N histograms which covers the area which is computed
by multiplying given ratio and all the area of hisotgrams.
3. Compute minimum distance between a pixel and the N histograms for each pixel in the polygon.


garaemon:
test sometimes fails because of timeout.
error code is 408.

http://qa.atmarkit.co.jp/q/1014

PR #1155, opened 2015-09-05
add yaml-cpp

k-okada:
to fix http://jenkins.ros.org/view/IbinS32/job/ros-indigo-jsk-recognition-utils_binarydeb_saucy_i386/


garaemon:
thanks

PR #1170, opened 2015-09-07
Facilitate visualization of README

wkentaro:


wkentaro:
see: https://github.com/wkentaro/jsk_recognition/blob/cleanup-readme/README.md

PR #999, opened 2015-07-07
[jsk_perception] add cmt_nodelet depending on libcmt

aginika:
DON'T MERGE YET MANY POINTS TO REPAIR.

libcmt_test

PR #1186, opened 2015-10-02
[jsk_perception] Add CMTNodelet, updated from #999

garaemon:


garaemon:
libcmt is not released on hydro because hydro packages are no longer updated.
Therefore, hydro should fail.

PR #1197, opened 2015-10-06
[jsk_recognition_utils] Fix camera sensor model

garaemon:


garaemon:
It was correct

screenshot from 2015-10-06 17 41 36


screenshot from 2015-10-06 17 41 52

PR #1226, opened 2015-10-14
[jsk_pcl_ros] Increase max value of max_size for EuclideanClustering

wkentaro:


wkentaro:
Maybe max_size should be ratio to size of all points when keep organized.

garaemon:
> Maybe max_size should be ratio to size of all points when keep organized.

It depends on context.

wkentaro:
So is that possible to change it to allow float value. 
* float value -> ratio
* int value -> number of point

garaemon:
> So is that possible to change it to allow float value? like:

We have no good answer for it.
number of points changes according to pre-processing and sensor model.
size of each point changes according to distance (and sensor model).

I'd like to keep simple number to filter out small clusters in order to avoid these discussion.

You can implement rejection nodelet according to your context.

PR #1231, opened 2015-10-14
[jsk_pcl_ros] Convert cluster point indices to mask and label image

wkentaro:
cpi_to_mask_image
cpi_to_label

### DIFF
--- a/jsk_pcl_ros/src/cluster_point_indices_decomposer_nodelet.cpp
+++ b/jsk_pcl_ros/src/cluster_point_indices_decomposer_nodelet.cpp
@@ -34,6 +34,7 @@
  *********************************************************************/
 
 #include "jsk_pcl_ros/cluster_point_indices_decomposer.h"
+#include <cv_bridge/cv_bridge.h>
 #include <pluginlib/class_list_macros.h>
 #include <pcl/filters/extract_indices.h>
 #include <pcl/common/centroid.h>
@@ -42,6 +43,7 @@
 #include <pcl/registration/ia_ransac.h>
 #include <pcl/filters/project_inliers.h>
 #include <pcl/common/pca.h>
+#include <sensor_msgs/image_encodings.h>
 #include <jsk_topic_tools/color_utils.h>
 #include <Eigen/Geometry> 
 
@@ -72,6 +74,7 @@ namespace jsk_pcl_ros
     pnh_->param("force_to_flip_z_axis", force_to_flip_z_axis_, true);
     pc_pub_ = advertise<sensor_msgs::PointCloud2>(*pnh_, "debug_output", 1);
     box_pub_ = advertise<jsk_recognition_msgs::BoundingBoxArray>(*pnh_, "boxes", 1);
+    mask_pub_ = advertise<sensor_msgs::Image>(*pnh_, "mask", 1);
   }
 
   void ClusterPointIndicesDecomposer::subscribe()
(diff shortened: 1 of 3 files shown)


garaemon:
only for organized pointcloud?

jsk_pcl_ros/src/cluster_point_indices_decomposer_nodelet.cpp
```diff
@@ -361,6 +367,14 @@ namespace jsk_pcl_ros
                                                input->header.frame_id,
                                                tf_prefix_ + (boost::format("output%02u") % (i)).str()));
       }
+      // create mask & label image from cluster indices
+      for (size_t j = 0; j < sorted_indices[i]->size(); j++) {
+        int index = sorted_indices[i]->data()[j];
+        int width_index = index % input->width;
+        int height_index = index / input->width;
+        mask.at<uchar>(height_index, width_index) = 255;
+        label.at<int>(height_index, width_index) = (int)i;
```
garaemon: 0 should be avoided

wkentaro:
> only for organized pointcloud?

That only use point indices, so I think organized/unorganized is not important.
organized/unorganized is important when we process point cloud, right?

jsk_pcl_ros/src/cluster_point_indices_decomposer_nodelet.cpp
wkentaro: > 0 should be avoided

I will update. thanks.

wkentaro:
I updated the commit.

garaemon:
conversion to label image only works for image-like organized pointcloud

need to verify it

2015年10月15日木曜日、Kentaro Wada<notifications@github.com>さんは書きました:

> I updated the commit.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_recognition/pull/1231#issuecomment-148132470>
> .
>


-- 
✉︎ from iPhone

wkentaro:
> conversion to label image only works for image-like organized pointcloud

It uses input cloud's width and height value.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1231/files#diff-e9363841a8ce4af3c4e097a96239ef4fR341

If it is NULL (maybe the pointcloud sensed by laser has NULL), the label and mask should be disabled.
How can I verify it? Should I check if it is `int` or are there some special value?
Actually I don't know what kind of value is there when sensed with laser.

garaemon:
non organized pointcloud has hesght=1 and generated images should be invalid

wkentaro:
Ok. I will skip creating mask & label image when height is 1.

wkentaro:
@garaemon 
is this change ok? 
https://github.com/wkentaro/jsk_recognition/commit/e11e36a3302560e0c7ee907f0966766dee2d2aa8?w=1

garaemon:
looks good

2015年10月15日木曜日、Kentaro Wada<notifications@github.com>さんは書きました:

> @garaemon <https://github.com/garaemon>
> is this change ok?
> wkentaro@e11e36a?w=1
> <https://github.com/wkentaro/jsk_recognition/commit/e11e36a3302560e0c7ee907f0966766dee2d2aa8?w=1>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_recognition/pull/1231#issuecomment-148269489>
> .
>


-- 
✉︎ from iPhone

wkentaro:
I squashed the commits.
Please review after the test passes.

PR #1248, opened 2015-10-16
Add document about how to avoid compilation error because of memory starvation

garaemon:
@orikuma @terasawa


garaemon:
@eisoku9618

PR #1294, opened 2015-10-29
[jsk_pcl_ros] Add urdf voxel model generator

mmurooka:
generate_urdf

This PR include https://github.com/jsk-ros-pkg/jsk_recognition/pull/1293 because depending on it.


mmurooka:
Next, I write script to add this model to gazebo,
and it's difficult to decide the package to locate a gazebo-related script.
(hrpsys_gazebo_*** is not desirable because those packages are installed only by HRP2/JAXON user.)

PR #1295, opened 2015-10-29
[jsk_pcl_ros] Add model to gazebo in model generation script

mmurooka:
This PR includes commits in https://github.com/jsk-ros-pkg/jsk_recognition/pull/1293 and https://github.com/jsk-ros-pkg/jsk_recognition/pull/1294 because of dependency on those PRs.

add_voxel_gazebo


mmurooka:
@garaemon , if there is no problem, cloud you merge?

jsk_pcl_ros/scripts/voxel_urdf_client.py
```diff
@@ -0,0 +1,54 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--gazebo', default=False)
+    parser.add_argument('--model', default=False)
+    parser.add_argument('--loop', default=True)
+    args = parser.parse_args(rospy.myargv()[1:])
+
+    marker_sub = rospy.Subscriber("input", Marker, markerCallback)
+    generate_srv = rospy.ServiceProxy("/generate_voxel_urdf", VoxelModelGenerate)
+
+    while args.loop:
```
garaemon: Can you implement as callback rather than main loop?

garaemon:
Can you implement it without generating a file?

mmurooka:
> Can you implement it without generating a file?

Do you mean inserting a model into gazebo without generating a urdf file?
I don't know how to do that..
The following is the usage of the command to spawn model into gazebo,
and we need a model file to spawn.

```
$ gzfactory spawn -h
This tool for spawning or deleting models into or from a running Gazebo simulation.

  gzfactory <spawn|delete> [options]
    spawn   : Spawn new model. Must specify a SDF model file.
    delete  : Delete existing model. Must specify model name.


Allowed options:
  -h [ --help ]           produce this help message
  -f [ --sdf ] arg        SDF model file.
  -w [ --world-name ] arg Name of Gazebo world.
  -m [ --model-name ] arg Model name.
  -x [ --pose-x ] arg     set model x position.
  -y [ --pose-y ] arg     set model y position.
  -z [ --pose-z ] arg     set model z positione.
  -R [ --pose-R ] arg     set model roll orientation in radians.
  -P [ --pose-P ] arg     set model pitch orientation in radians.
  -Y [ --pose-Y ] arg     set model yaw orientation in radians.
```

jsk_pcl_ros/scripts/voxel_urdf_client.py
mmurooka: moved to callback.
When `--loop` is false, I'd like to finish this script after generating a model once.

garaemon:
http://gazebosim.org/tutorials?tut=plugins_world

mmurooka:
In that case, I need to reimplement program by C++ with depending on gazebo.
I think that jsk_pcl_ros should not depend on gazebo, because compiling process might become more complicated and time-consuming.
Which package should I add? I don't want to add to hrpsys_gazebo_tutorials because it is not related to hrpsys.

And I'd like to confirm your intention.
Which program do you think should be removed from PR.
I added three program.
1. generate voxel from point cloud. (C++)
2. generate urdf file from voxel. (python)
3. insert urdf to gazebo (python)
You mean 1. should be kept, and 2., 3. should be removed, right?
And then, 2. and 3. program should be written as gazebo plugin, which no urdf file outputs.

Generating urdf itself is also meaningful because we can show the model in rviz as RobotModel.

garaemon:
> 1. generate voxel from point cloud. (C++)
> 2. generate urdf file from voxel. (python)
> 3. insert urdf to gazebo (python)
> You mean 1. should be kept, and 2., 3. should be removed, right?
> And then, 2. and 3. program should be written as gazebo plugin, in which no urdf file outputs.

1 sounds good.

2 sounds bad idea. 

3 is, I believe, already implemented in gazebo.

k-okada:
@mmurooka ping, 
 do we still need this type of feature ? ( i think yes), if so may be you can re-implement this or you'll find another way to archive this.

mmurooka:
> do we still need this type of feature ?

I don't need this right now, but maybe become to need some day.

> you can re-implement this or you'll find another way to archive this.

I'm still thinking that it is not bad to generate temporary urdf file to spawn recognized object to gazebo.
How do you think about that?

k-okada:
if purpose of this pr is to spawn octtree to the gazebo model, I think we can create spawn node to do that which we can change the shape on realtime, and that would be worth to put in upstream package,.

PR #1288, opened 2015-10-27
[jsk_perception] Test label image decomposer async

wkentaro:

PR #1302, opened 2015-11-05
Test commit for #1288

wkentaro:


wkentaro:
the problem on #1288 in travis test is resolved.

PR #1318, opened 2015-11-14
[jsk_pcl_ros] Add VoxelGridLargeScale

garaemon:


garaemon:
PCL's VoxelGrid cannot handle small voxel grids which may exceeds max value of int32.
VoxelGridLargeScale can handle it but it may takes long time to compute.

garaemon:
For example, multisense laser FoV exceeds 10m and if downsample pointcloud of multisense laser with 0.001 m voxel grid, it takes 10-20 seconds because VoxelGridLargeScale split spaces into about 800 subspaces.

PR #1172, opened 2015-09-10
[jsk_pcl_ros] Do not compile build_check.cpp in normal compilation time, just in run_tests

garaemon:
Little bit a tricky, do not compile build_check.cpp in normal compilation because ros buildfarm looks fail
to build it (memory starvation?).

http://jenkins.ros.org/view/IbinS64/job/ros-indigo-jsk-pcl-ros_binarydeb_saucy_amd64/69/console
```
[ 56%] Building CXX object CMakeFiles/build_check.dir/build_check.cpp.o
/usr/lib/ccache/c++   -DROSCONSOLE_BACKEND_LOG4CXX -DROS_PACKAGE_NAME=\"jsk_pcl_ros\" -DUSE_OLD_YAML -g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -DNDEBUG -D_FORTIFY_SOURCE=2   -Wno-deprecated -fopenmp -O2 -g -DNDEBUG -I/tmp/buildd/ros-indigo-jsk-pcl-ros-0.3.5-0saucy-20150910-0811/obj-x86_64-linux-gnu/devel/include -I/usr/include/vtk-5.8 -I/usr/include/opencv -I/tmp/buildd/ros-indigo-jsk-pcl-ros-0.3.5-0saucy-20150910-0811/include -I/opt/ros/indigo/include -I/usr/include/eigen3 -I/usr/include/pcl-1.7 -I/usr/include/ni -I/usr/include/qhull    -O2 -g -o CMakeFiles/build_check.dir/build_check.cpp.o -c /tmp/buildd/ros-indigo-jsk-pcl-ros-0.3.5-0saucy-20150910-0811/obj-x86_64-linux-gnu/build_check.cpp
Build timed out (after 120 minutes). Marking the build as aborted.
```

PR #1328, opened 2015-11-17
[jsk_pcl_ros] Test EuclideanSegmentation

wkentaro:


wkentaro:
catkin run_tests fails because of `build_check` does link `.so` file while testing. (if run test with multiple jobs)
but disabling build_check from run_tests is a problem https://github.com/jsk-ros-pkg/jsk_recognition/pull/1172 .
do you have idea? [at]garaemon

What build_check is doing? passing `catkin build` is not enough?

wkentaro:
This is problem. (`-j1` does work)
image

wkentaro:
I think [at]garaemon does not have time to correspond with this problem, so I make run_tests runs in a job. with https://github.com/wkentaro/jsk_recognition/commit/2f4befb533eb166b87651068c80f6a2cbd9f0ef5

wkentaro:
depends on https://github.com/jsk-ros-pkg/jsk_recognition/pull/1330

wkentaro:
Please merge.

PR #1330, opened 2015-11-18
Use gcc -z defs to check undefined symbols in shared objects (jsk_recognitoin_utils, jsk_pcl_ros, jsk_perception).

garaemon:
build_check.cpp cannot run on the environment using  multiple processes
because of invoking libjsk_pcl_ros.so link.


garaemon:
jsk_rviz_plugins, jsk_topic_tools should have "-z defs" option in linker flags too

wkentaro:
+1

PR #1341, opened 2015-11-24
add applying blur to output image on edge detector

YoheiKakiuchi:


k-okada:
please send PR to
https://github.com/ros-perception/vision_opencv/blob/indigo/opencv_apps/src/nodelet/edge_detection_nodelet.cpp,
when the function becomes stable

--
◉ Kei Okada

Issue #1384, opened 2015-12-09
Remove dependency to ml_classifiers


garaemon:
ml_classifiers is not maintained anymore.

garaemon:
```
$ rosdep what-needs ml_classifiers
jsk_pcl_ros
sklearn
$ rosdep what-needs sklearn       
jsk_pcl_ros
```

YutoUchimi:
`ml_classifiers` is already released in kintic, lunar and melodic.
See also https://github.com/jsk-ros-pkg/jsk_recognition/issues/2359

PR #1386, opened 2015-12-09
Remove Dependency to ml_classifiers

aginika:
#1384


garaemon:
I did not mean just removal of codes.

Can you rewrite these codes without ml_classifiers?

PR #1441, opened 2015-12-19
[jsk_perception] Fix tile_image.py for hydro.

garaemon:
1. Disable approximate sync for hydro. it's not supported on hydro
2. Use PIL.Image.frombytes instead of PIL.Image.fromstring

### DIFF
--- a/jsk_perception/node_scripts/tile_image.py
+++ b/jsk_perception/node_scripts/tile_image.py
@@ -9,7 +9,7 @@ from jsk_topic_tools import ConnectionBasedTransport
 import message_filters
 import rospy
 from sensor_msgs.msg import Image
-
+import os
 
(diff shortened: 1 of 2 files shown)


wkentaro:
`message_filters` supports approximate sync from 1.11.4. https://github.com/ros/ros_comm/commit/215ae253bab9dd1fce3e119cd1c0bb95a74a35f5
Why not check the version of that? You can get the version as following:
```python
import pkg_resources
from distutils.version import StrictVersion
if StrictVersion(pkg_resources.get_distribution('message_filters').version) < StrictVersion('1.11.4'):
  # here for old message_filters
  pass
else:
  # here for newer
  pass
```

garaemon:
done

PR #1490, opened 2016-01-09
 [jsk_pcl_ros] Add new feature to skip tracking according to     background substraction.

garaemon:
Sample launch is tabletop_tracking.launch

Now particle_filter_tracking can skip tracking when object looks stable
and difference pointcloud (which should be computed by
octree_change_detector)
are far from target object.

It can dramatically decrease computation.

screenshot from 2016-01-09 18 48 04


garaemon:
movie: https://drive.google.com/a/jsk.imi.i.u-tokyo.ac.jp/file/d/0B4bnoIcbRJVuci1WdW03ZkNFdnM/view?usp=sharing

PR #1499, opened 2016-01-21
Use jsk pcl ros utils namespace in jsk pcl nodelets

orikuma:
## Problem
When I run nodelets defined in jsk_pcl_ros_utils like colorize_distance_from_plane, nodelets abort with following errors. 
```
$ rosrun jsk_pcl_ros colorize_distance_from_plane
[ INFO] [1453269719.926632252]: Initializing nodelet with 4 worker threads.
[ERROR] [1453269720.344240963]: Failed to load nodelet [/colorize_distance_from_plane] of type [jsk_pcl/ColorizeDistanceFromPlane] even after refreshing the cache: MultiLibraryClassLoader: Could not create object of class type jsk_pcl_ros::ColorizeDistanceFromPlane as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
[ERROR] [1453269720.344310564]: The error before refreshing the cache was: MultiLibraryClassLoader: Could not create object of class type jsk_pcl_ros::ColorizeDistanceFromPlane as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
```
This error also occurs when  `rosrun nodelet nodelet standalone colorize_distance_from_plane` and nodelets are defined in launch files.
This error is similar to https://github.com/jsk-ros-pkg/jsk_recognition/issues/1482 , but in this case errors are reproduced whether I wait 60sec or not.

## Environments
This error occured in following environments.
- my laptop (ubuntu14.04 + indigo)
- hrp2007 auditor (ubuntu 12.04 + hydro)

However, this error did not occur in following environments.
- hrp2007v (ubuntu 12.04 + hydro)
- my desktop (ubuntu 12.04 + hydro)

In all environments, `catkin clean -a`, `git pull`  and `catkin bt` are executed in master branch of jsk_recognition.

## Possible solutions
I think this problem is caused by difference of namespace in which nodelets are defined.
In the jsk_pcl_ros_utils package, nodelets are defined in jsk_pcl_ros_utils namespace but in jsk_pcl_nodelets.xml of jsk_pcl_ros, these nodelets are defined as jsk_pcl_ros namespace.
Therefore, I fix this difference in this PR by following solutions:
- Use `jsk_pcl_ros_utils::Hoge` instead of `jsk_pcl_ros::Hoge` for type of nodelets in jsk_pcl_nodelets.xml in jsk_pcl_ros to correct namespace
- Use `jsk_pcl/Hoge` as name of nodelets for backward compatibility in jsk_pcl_ros
- Use `jsk_pcl_utils/Hoge` as name of nodelets in jsk_pcl_ros_utils to avoid collision of name. 
- Use `jsk_pcl_utils/Hoge` to pass single_nodelet_exec.cpp.in in jsk_pcl_ros_utils and `jsk_pcl/Hoge` in jsk_pcl_ros.

In hrp2007 auditor and my laptop, this PR solved above problem.
However, I don't have certainty for this PR and I want for this PR to be checked carefully by exparts.

## Remaining questions
I'm not sure why hrp2007v and my desktop can run `rosrun jsk_pcl_ros colorize_distance_from_plane` without this PR.
One possible answer is order of loading nodelets, which means if `jsk_pcl/Hoge` defined in jsk_pcl_ros_utils is loaded first, it is correctly loaded because its type is `jsk_pcl_ros_utils::Hoge` but if `jsk_pcl/Hoge` defined in jsk_pcl_ros is loaded first, it fails because its type is `jsk_pcl_ros::Hoge`.
However, I could not check this hypothesis because I do not know loading order of nodelets.

### DIFF
--- a/jsk_pcl_ros/jsk_pcl_nodelets.xml
+++ b/jsk_pcl_ros/jsk_pcl_nodelets.xml
@@ -1,155 +1,155 @@
 <library path="libjsk_pcl_ros_utils">
-  <class name="jsk_pcl/PCDReaderWithPose" type="jsk_pcl_ros::PCDReaderWithPose"
+  <class name="jsk_pcl/PCDReaderWithPose" type="jsk_pcl_ros_utils::PCDReaderWithPose"
          base_class_type="nodelet::Nodelet">
     <description>
       reconstract template pointcloud with recognized_pose
     </description>
   </class>
   <class name="jsk_pcl/AddPointIndices"
-         type="jsk_pcl_ros::AddPointIndices"
+         type="jsk_pcl_ros_utils::AddPointIndices"
          base_class_type="nodelet::Nodelet">
   </class>
   <class name="jsk_pcl/MaskImageToPointIndices"
-         type="jsk_pcl_ros::MaskImageToPointIndices"
+         type="jsk_pcl_ros_utils::MaskImageToPointIndices"
          base_class_type="nodelet::Nodelet">
   </class>
   <class name="jsk_pcl/MaskImageToDepthConsideredMaskImage"
-         type="jsk_pcl_ros::MaskImageToDepthConsideredMaskImage"
+         type="jsk_pcl_ros_utils::MaskImageToDepthConsideredMaskImage"
          base_class_type="nodelet::Nodelet">
     Generate depth considered mask image out of mask image
   </class>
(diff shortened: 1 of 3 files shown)


orikuma:
I fixed coments which is sent by mistake.

PR #1538, opened 2016-02-23
EuclidKClustering with reconfiguration of tolerance

wkentaro:


wkentaro:
https://github.com/start-jsk/jsk_apc/blob/master/jsk_2015_05_baxter_apc/node_scripts/euclid_k_clustering.py

PR #1550, opened 2016-03-20
remove dynamic_reconfigure.parameter_generator, which only used for rosbuild

k-okada:


k-okada:
related jsk-ros-pkg/jsk_common#1359

Issue #1558, opened 2016-03-21
.cfg files need to be fixed and released


dirk-thomas:
Recently you fixed some of the dynamic reconfigure files in this repo (b79ec0a1f8638a87aea774629e465cc38ef7bd94).Other files still need the same fix: e.g. https://github.com/jsk-ros-pkg/jsk_recognition/blob/edc9d417753fec6b1705aaa7ab83e6c888663649/jsk_perception/cfg/PolygonArrayColorLikelihood.cfg#L4-L10 https://github.com/jsk-ros-pkg/jsk_recognition/blob/edc9d417753fec6b1705aaa7ab83e6c888663649/jsk_pcl_ros_utils/cfg/CloudOnPlane.cfg#L6-L12

The fixes also need to be released. Currently the Jade jobs fail since the old code falls back to the non-catkin code generation which won't generate the header in the correct location.

This will likely also be the same for the upcoming Indigo rebuild.

k-okada:
Thanks, fixed in #1559

PR #1559, opened 2016-03-22
fix to compile on build server

k-okada:
ex. http://build.ros.org/job/Jbin_uU64__jsk_perception__ubuntu_utopic_amd64__binary/28/

- remove rosbuild from run/build depend
- remove dynamic_reconfigure.parameter_generator, which only used for rosbuild


k-okada:
Closes #1558

PR #1563, opened 2016-03-30
[jsk_perception] Add RectArrayActualSizeFilter

garaemon:
Filtering array of rectangle regions based on actual size estimated from
depth image.

rect_array_actual_size_filter


wkentaro:
@garaemon How do you draw the RectArray onto the Image?

wkentaro:
Ah, you use http://jsk-recognition.readthedocs.org/en/latest/jsk_perception/nodes/rect_array_to_image_marker.html ?
Or are there easier way to view it on image_view and rviz?

garaemon:
image_view does not support marker stuff and rviz is not good at visualizing 2D image processing.
We can make a simple program which takes an image and rect array and publish an image with drawing rectangles.

wkentaro:
I see. 

> We can make a simple program which takes an image and rect array and publish an image with drawing rectangles.

This has not yet done, have it?

garaemon:
> This has not yet done, have it?

Not yet done

wkentaro:
Ok, thank you!

wkentaro:
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1568

PR #1574, opened 2016-04-02
Stop using jenkins on travis testing

wkentaro:


wkentaro:
Now jenkins is restored.

PR #1585, opened 2016-04-11
[jsk_pcl_ros] add jsk_pcl version of tabletop_object_detector launch/config

furushchev:
also includes bug fix.


furushchev:
@k-okada @YoheiKakiuchi  Please review.

PR #1593, opened 2016-04-18
HOG modified and bug fix

iKrishneel:


k-okada:
please separate your PR into  bug fix and style change, and since we do not define style for our repository, if you start changing code style it will become unfinished task, so I'd suggest you to leave current style as it is, or change all code style and add lint test program.

iKrishneel:
Style change? I think i use 4 space indentation. That's similar to how this
file was initially.

k-okada:
please check you diffs again
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1593/files

--
◉ Kei Okada

k-okada:
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1593/files#diff-1767c4183b288c6b2d67af7a7b53dfdbL88

--
◉ Kei Okada

iKrishneel:
Ok I see. Will fix that. Besides, this PR is just for the descriptor which is used by 
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/src/sliding_window_object_detector.cpp
do I still need to add lint test.

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -105,29 +119,13 @@ cv::Mat HOGFeatureDescriptor::computeHOG(
        cv::cvtColor(image, image, CV_BGR2GRAY);
     }
     cv::Mat bins;
-    this->imageGradient(image, bins);
+    this->imageGradient(image, bins);
```
k-okada: spaces

jsk_perception/include/jsk_perception/histogram_of_oriented_gradients.h
```diff
@@ -1,44 +1,44 @@
...
+    HOGFeatureDescriptor(
+        const int = 8, const int = 2, const int = 9,
+        const float = 180.0f, const int = 1);
+    virtual cv::Mat computeHOG(const cv::Mat &);
+    template<typename T>
+    T computeHOGHistogramDistances(
+        const cv::Mat &, std::vector<cv::Mat> &,
+        const int = CV_COMP_BHATTACHARYYA);
+    cv::Mat visualizeHOG(
+        cv::Mat &, cv::Mat &, cv::Size, int, double);
```
k-okada: very hard to see which function did you added and and removed. did you add visualizeHOG() and update arguments of constructor? rename bilinearBinVoting to histgramBinVoting?

iKrishneel: Yes I did update the constructor to dynamic param setting. The visualizeHOG is  still not completed yet. Its just a HOG visualizer which I will add in later. 
-- Also  I changed is the name bilinearBinVoting to histgramBinVoting as I dont actually do bilinear. Instead to make it fast, I find one bin (lower)  while second one (higher) is just increment of lower. This saves computation mostly when used for tracking application

However major change is in this function blockGradient:
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1593/files#diff-1767c4183b288c6b2d67af7a7b53dfdbR85

k-okada: I prefer diff files that is easy to understand what did you
add/update/remove. your diff should basically look like

   HOGFeatureDetector(
+ const int = 8 ....
   )
- bilinearBinVoting
+ histgramBinVoting
+ visualizeHOG

--
◉ Kei Okada

iKrishneel: I see. There were few changes to almost all the functions. Besides the spacing, should I change something else.

k-okada: Not right now

--
◉ Kei Okada

2016/04/18 23:21、Krishneel <notifications@github.com> のメッセージ:

k-okada:
please update/add doc too -> https://github.com/jsk-ros-pkg/jsk_recognition/tree/master/doc/jsk_perception/nodes @iKrishneel

k-okada:
please update/add doc too -> https://github.com/jsk-ros-pkg/jsk_recognition/tree/master/doc/jsk_perception/nodes @iKrishneel

iKrishneel:
My apologies I totally overlooked this. I will update the PR with docs.

k-okada:
Ok, I have pushed a part of this commit, which I understand, still not clear the rest parts and added comments in the source,

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
+// Copyright (C) 2016 by Krishneel Chaudhary @ JSK Lab, The University
+// of Tokyo
 
 #include <jsk_perception/histogram_of_oriented_gradients.h>
 
-HOGFeatureDescriptor::HOGFeatureDescriptor() {
-   
+HOGFeatureDescriptor::HOGFeatureDescriptor(
+    const int cell_size, const int block_per_cell,
+    const int n_bins, const float angle, const int num_threads) :
```
k-okada: `num_threads` is not used in this Pull Request.

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
...
                 nearest_lower = distance;
                 lower_index = i;
             }
-        } else {
-            if (distance < nearest_higher) {
-                nearest_higher = distance;
-                higher_index = i;
-            }
         }
+	/*
```
k-okada: Does this never happens? or is this a bug?

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
...
 }
 
 void HOGFeatureDescriptor::imageGradient(
     const cv::Mat &image, cv::Mat &hog_bins) {
     cv::Mat xsobel;
     cv::Mat ysobel;
-    cv::Sobel(image, xsobel, CV_32F, 1, 0, 3);
-    cv::Sobel(image, ysobel, CV_32F, 0, 1, 3);
+    cv::Sobel(image, xsobel, CV_32F, 1, 0, 7);
+    cv::Sobel(image, ysobel, CV_32F, 0, 1, 7);
```
k-okada: Opencv sample code uses default param for sobel kernel size
https://github.com/opencv/opencv/search?utf8=%E2%9C%93&q=sobel+hog
and the default value is 3 -> http://docs.opencv.org/2.4/modules/imgproc/doc/filtering.html?highlight=sobel#sobel

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
...
     cv::Mat Imag;
     cv::Mat Iang;
     cv::cartToPolar(xsobel, ysobel, Imag, Iang, true);
-    cv::add(Iang, cv::Scalar(180), Iang, Iang < 0);
-    cv::add(Iang, cv::Scalar(-180), Iang, Iang >= 180);
+    cv::add(Iang, cv::Scalar(ANGLE), Iang, Iang < 0);
```
k-okada: Is this correct?  I do not understand what this code is, but if we change `cv::Scalar(0)` to `cv::Scalar(ANGLE)` then `Iang < 0` need to be changed too?

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
...
+		(rect.y + rect.height <= image.rows)) {
+		cv::Mat bin = cv::Mat::zeros(1, N_BINS, CV_32F);
+		for (int y = rect.y; y < (rect.y + rect.height); y++) {
+		    for (int x = rect.x; x < (rect.x + rect.width); x++) {
+			float angle = static_cast<float>(Iang.at<float>(y, x));
+			int l_bin;
+			int h_bin;
+			this->histogramBinVoting(angle, l_bin, h_bin);
+			float l_ratio = 1.0f - (angle - l_bin)/BINS_ANGLE;
+			float h_ratio = 1.0f - l_ratio;
```
k-okada: This is changed from `float h_ratio = 1 + (angle - h_bin)/BINS_ANGLE;`

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
...
-          int index = (j * CELL) + i;
-          for (int k = 0; k < N_BINS; k++) {
-             block_hogMD.at<float>(0, icounter++) = bins.at<float>(index);
-          }
-        }
+    for (int j = 0; j < BLOCK; j++) {
+        for (int i = 0; i < BLOCK; i++) {
+	    int index = i + (j * stride) + col;
+	    for (int k = 0; k < N_BINS; k++) {
+		block_hogMD.at<float>(0, icounter++) = bins.at<float>(index, k);
```
k-okada: this part is changed from 
```
-             block_hogMD.at<float>(0, icounter++) = bins.at<float>(index);

```

jsk_perception/src/histogram_of_oriented_gradients.cpp
```diff
@@ -1,99 +1,112 @@
...
-                (rect.y + rect.height <= image.rows)) {
-               cv::Mat hogMD = this->blockGradient(rect.x, rect.y, bins);
-               normalize(hogMD, hogMD, 1, 0, CV_L2);
-               featureMD.push_back(hogMD);
-            }
+    const int stride = static_cast<int>(image.cols/CELL);
+    int index = 0;
+    for (int j = 0; j < image.rows - CELL; j += CELL) {
+	for (int i = 0; i < image.cols - CELL; i += CELL) {
+	    cv::Mat hogMD = this->blockGradient(index, index, stride, bins);
```
k-okada:
```
-            if ((rect.x + rect.width <= image.cols) &&
 -                (rect.y + rect.height <= image.rows)) {
```

k-okada:
btw do we have opencv sample code hosted at https://github.com/opencv/opencv/tree/master/samples

jsk_perception/src/histogram_of_oriented_gradients.cpp
iKrishneel: In this PR, there was no Multi Threading. Needs to be deleted

jsk_perception/src/histogram_of_oriented_gradients.cpp
iKrishneel: The idea is to bi-linearly interpolate, so if some distribution goes into lower bin, then the upper bin is just lower_bin+1. Save  searching time.

jsk_perception/src/histogram_of_oriented_gradients.cpp
iKrishneel: Yes that is correct, but Sobel operator with higher kernel can give better response..

jsk_perception/src/histogram_of_oriented_gradients.cpp
iKrishneel: Basically, if the angle is less than 0, just add 180. Likewise if the angle is greater than 180, just subtract by 180.

jsk_perception/src/histogram_of_oriented_gradients.cpp
iKrishneel: Same as above where we just use lower_bin_index + 1 = higher_bin_index.

jsk_perception/src/histogram_of_oriented_gradients.cpp
iKrishneel: just a condition that I feel is not necessary .

iKrishneel:
This code I wrote it for tracking, so it was optimized for time and the purpose.
Incase if someone once to use HOG for detection the OpenCV implementation [cuda::HOG](http://docs.opencv.org/3.0-beta/modules/cuda/doc/object_detection.html) should be better.

k-okada:
commit the part of this PR in #1863 
You'd better to add new function named ` void HOGFeatureDescriptor::bilinearBinVoting(` and keep old one as `void HOGFeatureDescriptor::histogramBinVoting(` so that we can back to original version if necessary, or add test code that ensure changed code is same (or better) than current one.
If you use cuda::HOG, you can use https://github.com/opencv/opencv/blob/master/samples/gpu/hog.cpp and we're working project to ros-nize opencv sample to ros package at https://github.com/ros-perception/opencv_apps/blob/indigo/CMakeLists.txt#L105
I personally think it is very important to compare compare your work against old one, like bilinear vs histogram, not update code just by 'I feel this works better' and if you find something you can find a material to write on your paper, and as for optimizing for time, if that speed up does change any situation, that's ok, but usually, if you downsample the imput data, you can get more speed. So it is important to think what can we get by speeding up the algorithm.

PR #1601, opened 2016-04-22
[jsk_recognition_utils] Stop passing -z flag to ld with Clang

wkentaro:


jsk_recognition_utils/CMakeLists.txt
```diff
@@ -42,7 +42,9 @@ IF(${yaml_cpp_VERSION} VERSION_LESS "0.5.0")
 ## indigo yaml-cpp : 0.5.0 /  hydro yaml-cpp : 0.3.0
   add_definitions("-DUSE_OLD_YAML")
 ENDIF()
-set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -z defs")
+if(NOT ${APPLE})
```
furushchev: @wkentaro I think this would be better to define condition by compiler instead of Operating System, because this option seems to be valid only for GCC compiler. As you may know, Apple uses clang instead of GCC as a default compiler and aliases it to GCC for backward compatibility.

wkentaro: @furushchev Thanks, I updated the commit.

PR #1606, opened 2016-04-23
[jsk_pcl_ros] Stop passing -z flag to ld with clang

wkentaro:


k-okada:
btw what was the problem for this

wkentaro:
There is "unknown option -z is passed error" with clang.

wkentaro:
and build fails.

PR #1609, opened 2016-04-23
[jsk_pcl_ros] Support pcl 1.8 in 'jsk_pcl_ros'

wkentaro:


k-okada:
can you add travsi test for this?

wkentaro:
In that case, we need to compile all the dependencies of jsk_pcl_ros from source, right?

wkentaro:
I added a before_script.

PR #1624, opened 2016-04-28
[jsk_pcl_ros/PointcloudDatabaseServer] fix and improve for frame_id

knorth55:
MERGE AFTER #1625 
Before modification, there is no string (such as `""`) in frame_id.
In this PR, frame_id is named after`.pcd` file's name without extension `.pcd`


wkentaro:
Could you please add a test?

knorth55:
ok, which do you prefer? i think test is better.

wkentaro:
You can right both without taking so much time, can't you? I mean using `<include>` tag in test launch file.

knorth55:
i see

k-okada:
should we wait for test program?

knorth55:
that is better

knorth55:
please merge this

PR #1642, opened 2016-05-02
Remove imagesift dependency on jsk_perception

wkentaro:


k-okada:
need rebase, 
% hope we can use https://github.com/ros-perception/image_pipeline/pull/173 instead of staci virutal camera in the future

wkentaro:
Please merge.

PR #1657, opened 2016-05-07
[jsk_recognition_utils] Cythonize Non-maximum Supression (NMS) baseline

wkentaro:
For faster computing while object detection.


wkentaro:
Please merge this.

k-okada:
if you need speedup, better to write in C++?

PR #1679, opened 2016-05-20
[jsk_pcl_ros] add approximate_sync parameter to depth_error.launch 

kochigami:
If I set approximate_sync parameter to true, ```/depth_error_image/output``` is published.
I tried it with xtion pro (ID 1d27:0601 ASUS), ubuntu 14.04, ros indigo.
I followed [this tutorial](https://jsk-recognition.readthedocs.io/en/latest/jsk_pcl_ros/calibration.html) to try this program .

I don't know whether setting parameter on roslaunch is better or adding information on doc is better.
If there is any problem, please let me know.

### DIFF
--- a/jsk_pcl_ros/launch/depth_error.launch
+++ b/jsk_pcl_ros/launch/depth_error.launch
@@ -39,6 +39,7 @@
     <remap from="~image" to="$(arg DEPTH_IMAGE)" />
     <remap from="~point" to="/checkerdetector/corner_point" />
     <remap from="~camera_info" to="$(arg CAMERA_INFO_TOPIC)" />
+    <param name="approximate_sync" value="true"/>
   </node>
 
 </launch>
--- a/jsk_pcl_ros/launch/depth_error.launch
+++ b/jsk_pcl_ros/launch/depth_error.launch
@@ -39,7 +39,9 @@
     <remap from="~image" to="$(arg DEPTH_IMAGE)" />
     <remap from="~point" to="/checkerdetector/corner_point" />
     <remap from="~camera_info" to="$(arg CAMERA_INFO_TOPIC)" />
-    <param name="approximate_sync" value="true"/>
+    <rosparam> 
+      approximate_sync: true
+    </rosparam>
   </node>
 
 </launch>


jsk_pcl_ros/launch/depth_error.launch
```diff
@@ -39,6 +39,7 @@
     <remap from="~image" to="$(arg DEPTH_IMAGE)" />
     <remap from="~point" to="/checkerdetector/corner_point" />
     <remap from="~camera_info" to="$(arg CAMERA_INFO_TOPIC)" />
+    <param name="approximate_sync" value="true"/>
```
k-okada: what happens if we use "~approximate_sync"? is this working or not?

kochigami:
I understood setting approximate_sync as using inputs which have matching timestamp.
=> http://wiki.ros.org/message_filters#ApproximateTime_Policy
However, I'm not still certain what is happened...

As for setting param, I referred to this, but I think it's better to use ```rosparam```. Sorry for my mistake.
=> http://wiki.ros.org/roslaunch/XML/param

I'll try calib program suceeds or not according on these conditions when I borrow xtion from someone, and then I'll update this or document:
- <rosparam> approximate_sync: true
- set nothing 
- <rosparam> approximate_sync: false

kochigami:
Sorry for my late response, but I confirmed that setting ```approcimate_sync``` to true has an effect.
I added it like this:
```
    <rosparam>
      approximate_sync: true
    </rosparam>
```

- set ```approximate_sync``` to false ,or writing nothing about ```approximate_sync``` to depth_error.launch:

```rosrun jsk_pcl_ros depth_error_calibration.py --model quadratic-uv-quadratic-abs``` returned only the background image of graph and did nothing
```rosrun image_view image_view image:=/depth_error_logistic_regression/frequency_image``` shows nothing

- set ```approximate_sync```  to true:

```rosrun jsk_pcl_ros depth_error_calibration.py --model quadratic-uv-quadratic-abs``` returns plot graph according to checkerboard detector input
```rosrun image_view image_view image:=/depth_error_logistic_regression/frequency_image``` returns plot graph

However, I don't know whether this situation is compatible for another PC.
I'll close this pull request and add information to tutorial.

PR #1628, opened 2016-04-29
[jsk_perception] Download trained_data with jsk_data function

wkentaro:

PR #1685, opened 2016-05-21
Upgrade jsk_travis to 0.4.8 and fix some regressions

wkentaro:
Closes #1684

PR #1686, opened 2016-05-21
Remove imagesift/CATKIN_IGNORE which is wrongly commited on #1628

wkentaro:


wkentaro:
Will be merged in #1685

PR #1687, opened 2016-05-21
[jsk_pcl_ros_utils] add PCD Capture nodelet

knorth55:
it captures pointcloud topic and save it into pcd file.

### DIFF
--- a/doc/jsk_pcl_ros_utils/nodes/pcd_capture.rst
+++ b/doc/jsk_pcl_ros_utils/nodes/pcd_capture.rst
@@ -1,0 +1,25 @@
+PCDCapture
+==========================
+
+What is this?
+-------------
+
+Capture point cloud topic and save in pcd file.
+
+Subscribing Topic
+-----------------
+
+* ``~input`` (``sensor_msgs/PoinCloud2``)
+
+  PointCloud.
+
+Parameters
+----------
+
+* ``filename`` (String, default: ``tmp.pcd``)
+  
+  PCD filename.
+
+* ``duration`` (Int, default: ``1``)
+
+  Saving duration. You can chage saving frequency with this parameter.
--- a/jsk_pcl_ros_utils/CMakeLists.txt
+++ b/jsk_pcl_ros_utils/CMakeLists.txt
@@ -139,6 +139,7 @@ generate_dynamic_reconfigure_options(
 #   cfg/MovingLeastSquareSmoothing.cfg
 #   cfg/OctreeVoxelGrid.cfg
 #   cfg/OctreeChangePublisher.cfg
+  cfg/PCDCapture.cfg
   )
 
 generate_messages(DEPENDENCIES
@@ -240,6 +241,8 @@ jsk_pcl_util_nodelet(src/pcd_reader_with_pose_nodelet.cpp
   "jsk_pcl_utils/PCDReaderWithPose" "pcd_reader_with_pose")
 jsk_pcl_util_nodelet(src/pointcloud_relative_from_pose_stamped_nodelet.cpp
   "jsk_pcl_utils/PointCloudRelativeFromPoseStamped" "pointcloud_relative_from_pose_stamped")
+jsk_pcl_util_nodelet(src/pcd_capture_nodelet.cpp
+  "jsk_pcl_utils/PCDCapture" "pcd_capture")
 add_library(jsk_pcl_ros_utils SHARED ${jsk_pcl_util_nodelet_sources})
 add_dependencies(jsk_pcl_ros_utils ${PROJECT_NAME}_gencpp ${PROJECT_NAME}_gencfg)
 target_link_libraries(jsk_pcl_ros_utils
--- a/jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
+++ b/jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
@@ -1,0 +1,99 @@
+// -*- mode: c++; indent-tabs-mode: nil; -*-
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2016, JSK Lab
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/o2r other materials provided
+ *     with the distribution.
+ *   * Neither the name of the JSK Lab nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+#define BOOST_PARAMETER_MAX_ARITY 7
+
+#include "jsk_pcl_ros_utils/pcd_capture.h"
+#include "jsk_recognition_utils/pcl_conversion_util.h"
+#include <pcl/io/io.h>
+#include <pcl/point_types.h>
+#include <pcl/exceptions.h>
+#include <sensor_msgs/PointCloud2.h>
+
+namespace jsk_pcl_ros_utils
+{
+  void PCDCapture::onInit()
+  {
+    DiagnosticNodelet::onInit();
+    srv_ = boost::make_shared <dynamic_reconfigure::Server<Config> > (*pnh_);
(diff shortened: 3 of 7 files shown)


knorth55:
I'm not sure whether i should make this function with rosservice or dynamic_reconfigure, but, in my case, dynamic_reconfigure is easier to handle.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
```diff
@@ -0,0 +1,99 @@
...
+    dynamic_reconfigure::Server<Config>::CallbackType f = boost::bind (
+        &PCDCapture::configCallback, this, _1, _2);
+    srv_->setCallback (f);
+    sub_ = pnh_->subscribe("input", 1, &PCDCapture::callback, this);
+    onInitPostProcess();
+  }
+
+  void PCDCapture::subscribe()
+  {
+  }
```
knorth55: I could not find the reason, but, in this nodelet, it never get into `subscribe` func so that i put subscribe in `onInit`.
If someone find wrong or missing point, please tell me. thank you

wkentaro:
Why don't you use [pointcloud_to_pcd](http://wiki.ros.org/pcl_ros#pointcloud_to_pcd) in pcl_ros?

knorth55:
because i want to use dynamic reconfigure

wkentaro:
> because i want to use dynamic reconfigure

for what?

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  void PCDCapture::callback(
+    const sensor_msgs::PointCloud2::ConstPtr& cloud_msg)
+  {
+    pcl::PointCloud<pcl::PointXYZRGB> cloud;
+    pcl::fromROSMsg(*cloud_msg, cloud);
+    try
+    {
+      boost::filesystem::path path (filename_);
+      ext_ = path.extension().string();
```
wkentaro: Need mutex lock if you change property.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
wkentaro: That is because this node will never be subscribed.
The parent class of DiagnosticNodelet is ConnectionBasedNodelet, and it subscribes only when being subscribed.
You don't need to use them in this case.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: which one?

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
```diff
@@ -0,0 +1,99 @@
...
+        filename_ += std::string(".pcd"); 
+      }
+      pcl::io::savePCDFileASCII (filename_, cloud); 
+      JSK_NODELET_INFO("pcd file saved: %s", path.c_str());
+    }
+    catch (pcl::IOException& e)
+    {
+      JSK_NODELET_ERROR("Failed to save pcd file");
+    }
+    ros::Duration(duration_).sleep();
```
wkentaro: Sleeping in callback is not good practice.
Use timer for saving.

doc/jsk_pcl_ros_utils/nodes/pcd_capture.rst
```diff
@@ -0,0 +1,25 @@
...
+-----------------
+
+* ``~input`` (``sensor_msgs/PoinCloud2``)
+
+  PointCloud.
+
+Parameters
+----------
+
+* ``filename`` (String, default: ``tmp.pcd``)
```
wkentaro: Filename should be like `{TIMESTAMP}.pcd` or `tmp_{TIMESTAMP}.pcd`.

jsk_pcl_ros_utils/CMakeLists.txt
```diff
@@ -240,6 +241,8 @@ jsk_pcl_util_nodelet(src/pcd_reader_with_pose_nodelet.cpp
   "jsk_pcl_utils/PCDReaderWithPose" "pcd_reader_with_pose")
 jsk_pcl_util_nodelet(src/pointcloud_relative_from_pose_stamped_nodelet.cpp
   "jsk_pcl_utils/PointCloudRelativeFromPoseStamped" "pointcloud_relative_from_pose_stamped")
+jsk_pcl_util_nodelet(src/pcd_capture_nodelet.cpp
+  "jsk_pcl_utils/PCDCapture" "pcd_capture")
```
wkentaro: why not `pointcloud_to_pcd`?

wkentaro:
In my opinion, it is fine to add similar node in jsk_pcl_ros as pcl_ros, because:

- The GitHub flow in [perception_pcl](https://github.com/ros-perception/perception_pcl) is too slow. (ex https://github.com/ros-perception/perception_pcl/pull/112)
- The source tree of pcl_ros is complicated, and timestamp is wrong in some nodes.

cc @k-okada 

but we need to follow name convention and possibility to be merged in upstream.
(maybe you can copy the pointcloud_to_pcd.cpp to jsk_pcl_ros then edit as you like)

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
wkentaro: `ext_` is the property, isn't it?

wkentaro: `filename_` also.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
wkentaro: This maybe prevents reconfigure event from entering the callback.

knorth55:
actually i didn't realize pcl_ros has same node when i wrote this.
but now i read pcl_ros source code, and i thought it is unease to use.
first, ii want to distinguish pcd files with names.
second, pcl_ros one save in every time the node receive point cloud, and i thought it is too much.
the last is that i want to turn on/off the saving function, but i'm not sure whether dynamic recofigure or rosservice is better.

knorth55:
i will change its name.

wkentaro:
> first, ii want to distinguish pcd files with names.

why? it saves point cloud as `{prefix}{TIMESTAMP}.pcd`

knorth55:
i want to change prefix with dynamic reconfigure

wkentaro:
> second, pcl_ros one save in every time the node receive point cloud, and i thought it is too much.

you can just add duration option, can't you?

wkentaro:
> i want to change prefix with dynamic reconfigure

I'm commenting in the context with below comment I described above.

> (maybe you can copy the pointcloud_to_pcd.cpp to jsk_pcl_ros then edit as you like)

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: i don't understand whay the property is. is it equal to global variable?

wkentaro:
> the last is that i want to turn on/off the saving function, but i'm not sure whether dynamic recofigure or rosservice is better.

rosservice is better I think. as suggested by vision_opencv maintainer in https://github.com/ros-perception/image_pipeline/pull/163

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: ok, thank you.
every nodelet has same idea?

wkentaro: > every nodelet has same idea?

?

wkentaro: FYI: http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/overview.html#nodelet-based-designe

knorth55:
> you can just add duration option, can't you?

of course, i can.

> I'm commenting in the context with below comment I described above.

what do you mean?

> rosservice is better I think. as suggested by vision_opencv maintainer in ros-perception/image_pipeline#163

thank you, i will follow that.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
wkentaro: slots in euslisp.

wkentaro:
> I'm commenting in the context with below comment I described above.
what do you mean?

I agree with adding this node, but disagree with idea of not reusing code in pcl_ros.
I understood your comment below as the latter.

> but now i read pcl_ros source code, and i thought it is unease to use.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: i think they are not properties...

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: thanks

knorth55:
>I agree with adding this node, but disagree with idea of not reusing code in pcl_ros.
I understood your comment below as the latter.

ok

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
wkentaro: You're right. according to the definition in [wikipedia](https://en.wikipedia.org/wiki/Property_(programming)), it is called as field or data member.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: > The parent class of DiagnosticNodelet is ConnectionBasedNodelet, and it subscribes only when being subscribed.

i wonder that the basic design of nodelet also have this idea.
i read ROS wiki, and it looks it is not a basic idea.

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
knorth55: do i need to lock these?

wkentaro: Yes, as well as in the call back for reconfigure.

wkentaro:
> ok, i understand what you say.
but is it related to prefix?

I recongnize you are not reusing the code from below conversation, so I quoted the last comment.

you> but now i read pcl_ros source code, and i thought it is unease to use.
         first, ii want to distinguish pcd files with names.
I > why? it saves point cloud to {prefix}{TIMESTAMP}.pcd
you> i want to change prefix with dynamic reconfigure

I recognized the last sentence as "i want to change prefix with dynamic reconfigure (so the code cannot be reused)".

jsk_pcl_ros_utils/src/pcd_capture_nodelet.cpp
wkentaro: > i read ROS wiki,

Where?

wkentaro: > i wonder that the basic design of nodelet also have this idea.

No, this is not the basic design of nodelet.
So you don't have to follow it.

wkentaro: > i wonder that the basic design of nodelet also have this idea.

However, authors of perception nodes usually use the connection based processing.
https://github.com/ros-perception/perception_pcl/blob/kinetic-devel/pcl_ros/src/pcl_ros/filters/filter.cpp#L178-L180

knorth55:
sorry for not telling you enough information.
i close this PR and remake new PR #1689.

PR #1709, opened 2016-05-30
Implement OverlayImageColorOnMono: nodelet that overlay color image on mono image

wkentaro:
cc @yuyu2172


wkentaro:
Please merge this.

k-okada:
@yuyu2172 Could you update sample image ? I think you have nice sample image with object recognition (over lay posterity image over input image)

yuyu2172:
target_posterior_image_in_bin

Float image represents how likely pixels are assigned to the category 'ball'.

k-okada:
very nice image, but please update sample image, but we want to remove left 'start bar' ,  you can choose 'grab current screen option' or move the rqt image viewer and grub again
hoge

k-okada:
nice, but does the image is upside down?

--
◉ Kei Okada

k-okada:
also bettor to remove mask from image, because this is to explain overlay,
not mask function.

--
◉ Kei Okada

yuyu2172:
@k-okada 
Probability likelihood of the object is calculated only in masked region, so we can only use masked image for this example.

wkentaro:
> @k-okada 
Probability likelihood of the object is calculated only in masked region, so we can only use masked image for this example.

In that case, why not overlay the posterior image on the masked color image?

yuyu2172:
that makes sense

yuyu2172:
screenshot from 2016-05-31 16 21 11

wkentaro:
1. Please add an image on which the color image is overlaid
2. Please update the topic name.

yuyu2172:
screenshot from 2016-05-31 17 32 46

yuyu2172:
screenshot from 2016-05-31 17 35 58

k-okada:
!!!! Great!

Can you add more example? Since this example shows red ball which seems very easy to recognize simple color filter so I'm interested in more difficult target

k-okada:
!!!! Great!

Can you add more example? Since this example shows red ball which seems very easy to recognize simple color filter so I'm interested in more difficult target

k-okada:
Can you recognize the book on left side?

yuyu2172:
Likelihood image for the book is not as clear as the image for the balls.
I think that the image for the balls is a good example for this node.

screenshot from 2016-05-31 17 35 58

yuyu2172:
The picture above is just an internal state of the segmentation algorithm.
Segmented 


frame0000

wkentaro:
Please merge this.

PR #1713, opened 2016-05-30
[jsk_pcl_ros] Fixed initialization of pnh in organized_edge_detector

iory:
in organized_edge_detector, 
we must initialize pnh, etc...


wkentaro:
LGTM

PR #1720, opened 2016-06-01
[jsk_pcl_ros] Fixed mistake of condition in edge_depth_refinment

iory:
condition max_x_index != max_x_index is always false


jsk_pcl_ros/src/edge_depth_refinement_nodelet.cpp
```diff
@@ -143,7 +143,7 @@ namespace jsk_pcl_ros
       }
     }
 
-    if (max_x_index != max_x_index) {
+    if (min_x_index != max_x_index) {
       return boost::make_tuple(
         max_x_index, min_x_index);
```
wkentaro: Why is it in this order?
Here it is `max, min`, but `min, max` [below](https://github.com/jsk-ros-pkg/jsk_recognition/pull/1720/files#diff-5475a8fb677341120b99f38181eb936aR164).

jsk_pcl_ros/src/edge_depth_refinement_nodelet.cpp
```diff
@@ -143,7 +143,7 @@ namespace jsk_pcl_ros
       }
```
wkentaro:
```c++
      if (y > max_y) {
         // max_y = max_y;
         max_y = y;
         max_y_index = index;
       }
```

iory: I overlooked this

wkentaro:
IMO, we should add a simple test if we change the existing program.

iory:
Ok, I will write test code.

iory:
Please merge this.

PR #1722, opened 2016-06-01
[doc/jsk_pcl_ros] Add documentation for edge_depth_refinement

iory:
For #1631


doc/jsk_pcl_ros/nodes/edge_depth_refinement.rst
```diff
@@ -0,0 +1,42 @@
+EdgeDepthRefinement
+=====================
```
wkentaro: Should be same length as text.

iory: I missed. Thanks!

wkentaro:
+1

PR #1733, opened 2016-06-07
[jsk_perception/solidity_rag_merge.py] fix bug in solidity_rag_merge

knorth55:


k-okada:
please add the error message and/or wrong output if we do not have this PR,  @knorth55

PR #1760, opened 2016-06-22
[jsk_perception/src/virtual_camera_mono.cpp] process only when subscribed

furushchev:


furushchev:
@k-okada Please merge if there is no problem.

furushchev:
@k-okada Please merge if there is no problem!

furushchev:
@k-okada Thanks!

PR #1770, opened 2016-07-04
Add line length limit

iory:
There was no effect of min_length_ parameter.
enabled min_length_ threshold and add test.


iory:
Please merge this.

PR #1772, opened 2016-07-05
Convert point indices to cluster point indices

wkentaro:
ex)

- Input Indices: [0, 10, 20]
- Output Cluster Indices: [[0, 10, 20]]


wkentaro:
Please merge this.

PR #1773, opened 2016-07-05
Convert point indices to mask w/o sync if it's static

wkentaro:
Because synchronization cause trouble sometimes, and in most cases image size is static.


wkentaro:
Please merge this.

PR #1775, opened 2016-07-05
[jsk_perception/fast_rcnn] Modified avoiding size of rects is 0 case

iory:
To deal with size of rects_orig is equal to 0, return from callback


wkentaro:
LGTM

iory:
Please merge this.

PR #1779, opened 2016-07-13
Adjust to opencv3

mizmizo:
Enabled to build with ros indigo, opencv-3.x.
It solves #1778


wkentaro:
Why not remove these lines to test this repository with OpenCV3 is installed?
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/.travis.yml#L26-L28

mizmizo:
That's accountable. thanks your advise.

k-okada:
Thank you , but in general you not have to use opencv3 until you really really really need this.

PR #1780, opened 2016-07-15
Convert point cloud to point indices

wkentaro:
## What is this?

- Move PointCloudToPointIndices to jsk_pcl_ros_utils from jsk_pcl_ros
- Accept unorganized point cloud

## Why?

- The node is utility and should be in jsk_pcl_ros_utils
- The node should accept unorganized point cloud also.


wkentaro:
**Please merge this!**

k-okada:
Does any of existing launch file need to be updated?

wkentaro:
> Does any of existing launch file need to be updated?

No, because the executable and nodelet declaration is still there.

k-okada:
may be https://github.com/jsk-ros-pkg/jsk_recognition/pull/1434 was not a good decision, we have to move files again and again...

wkentaro:
Is there another option?
Maybe spliting jsk_perception package will be required in the future.

PR #1797, opened 2016-07-29
Extract indices correctly with empty cloud

wkentaro:
Currently the node publishes point cloud with height=0, width=0 with empty point cloud.


wkentaro:
**Please merge this!**

PR #710, opened 2015-02-16
add imu_orientated plane detector and launch for icp-use

YuOhara:
icp_recognition for multisense and other_cameras.

PR #1799, opened 2016-07-29
Convert bounding box array to bounding box

wkentaro:
image


wkentaro:
**Please merge this!**

k-okada:
conflicts

wkentaro:
Please merge this.

mmurooka:
Cloud you tell me merits of implementing with C++ instead of using [python node](https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/scripts/box_array_to_box.py).

wkentaro:
Just I didn't know the existing node.
For this kind of light processing, I think there are few differences between implementation by C++ and Python.
Maybe the merits of the implementation in this PR are:

1. Node name is fit to standard: bounding_box_array_to_bounding_box.
2. The index can be changed with rosparam + dynamic reconfigure.
3. The node is connection-based, and it is efficient to reduce cpu load.
4. Topic names are fit to standard: BoundingBoxArray `~input` and BoundingBox `~output`.

mmurooka:
I see. Thank you.
> Do you have idea how to migrate it?

I don't know where `box_array_to_box.py` is already used, and it is difficult problem..

k-okada:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/scripts/box_array_to_box.py is cimmited by HRP2 user, who is this? @YuOhara, @YoheiKakiuchi, @orikuma, @terasawa

mmurooka:
maybe @YuOhara https://github.com/jsk-ros-pkg/jsk_recognition/pull/710

YuOhara:
I use box_array_to_box.py only in launch files in jsk_pcl_ros
and these will be changed to use c++ code in https://github.com/jsk-ros-pkg/jsk_recognition/pull/1833

k-okada:
BTW does C++ version has same interface as python one? i mean if i just
change the node name, does the launch file continue working?

--
◉ Kei Okada

wkentaro:
@k-okada 
Actually, **NO**. Because the default value of index for bounding box extraction from boxes is **-1** in the new nodelet, but **0** in the previous node.
So changes like below is necessary (setting rosparam `~index:=0` is also added)
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1833/files#diff-20f778dbf66e795624ef22069c9c08f9L55

PR #1803, opened 2016-07-30
[jsk_pcl_ros] Add drc_box_color2.yaml, which used in valve/door task demo but not commited

orikuma:


orikuma:
https://github.com/jsk-ros-pkg/jsk_demos/pull/1191 seems to work well in real jaxonred.

orikuma:
There were no problem in executing door and valve task sequentially on the real jaxonred

PR #1808, opened 2016-07-31
Compute point cloud centroid after transformed

wkentaro:
**Example** centroids computed in **BODY** frame


wkentaro:
**Please merge this!**

PR #1822, opened 2016-08-06
Color filter

iory:
Add 2d color filter.

hsi_color

### DIFF
--- a/jsk_perception/CMakeLists.txt
+++ b/jsk_perception/CMakeLists.txt
@@ -60,6 +60,8 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 endif()
 # Dynamic reconfigure support
 generate_dynamic_reconfigure_options(
+  cfg/RGBColorFilter.cfg
+  cfg/HSIColorFilter.cfg
   cfg/RectArrayActualSizeFilter.cfg
   cfg/PolygonArrayColorLikelihood.cfg
   cfg/PolygonArrayColorHistogram.cfg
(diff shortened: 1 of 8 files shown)


k-okada:
why not using
https://github.com/ros-perception/opencv_apps/blob/indigo/src/nodelet/camshift_nodelet.cpp
?

--
◉ Kei Okada

iory:
This is basic component of image-pipeline.

In Unix philosophy, it is said that following sentences.
```
Make each program do one thing well.
```
But camshift is aiming for tracking objects. it is 
this color filter does simplest work, filter color.

k-okada:
you can run 
```
roslaunch --screen opencv_apps camshift.launch image:=image_raw histogram:='[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255]' smin:=150 debug_view:=true 
```
and get labeled image at
```
rosrun image_view image_view image:=/camshift/back_project
```

we can also have rgb threshold opencv example so we can add this to oepncv_apps package.
https://github.com/opencv/opencv/blob/master/samples/cpp/tutorial_code/ImgProc/Threshold_inRange.cpp

iory:
OK!

mmurooka:
camshift can not set threshold for hue, so I'd like to use this program.
@iory , do you have plan to add this nodelet to opencv_apps?

mmurooka:
Oh, you already sent PR: https://github.com/ros-perception/opencv_apps/pull/48 . Please ignore my last message.

PR #1823, opened 2016-08-06
[jsk_perception] add warning message when no camera info is available.

mmurooka:
same with https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/src/polygon_array_to_label_image.cpp#L113-L115 .


mmurooka:
All tests passed. please merge if there is no problem.

PR #1825, opened 2016-08-06
[jsk_pcl_ros] add polygon_array_to_polygon.py

mmurooka:
This is polygon version of [box_array_to_box.py](https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/scripts/box_array_to_box.py).
I need this script to connect [OrganizedMultiPlaneSegmentation](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_pcl_ros/nodes/organized_multi_plane_segmentation.html) which publishes jsk_recognition_msgs/PolygonArray and [PolygonToMaskImage](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/polygon_to_mask_image.html) which subscribes geometry_msgs/PolygonStamped.

### DIFF
--- a/jsk_pcl_ros/scripts/polygon_array_to_polygon.py
+++ b/jsk_pcl_ros/scripts/polygon_array_to_polygon.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+import rospy
+
+PKG='jsk_pcl_ros'
+
+import imp
+## import message_filters
+try:
+    imp.find_module(PKG)
+except:
+    import roslib;roslib.load_manifest(PKG)
+
+from geometry_msgs.msg import PolygonStamped
+from jsk_recognition_msgs.msg import PolygonArray
+
+def polygon_array_cb(plolygon_array):
+    if (len(plolygon_array.polygons)==0):
+        rospy.loginfo("polygon size 0")
+        return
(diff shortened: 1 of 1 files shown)


wkentaro:
I have below suggestions:

1. Polygon itself is not 3D specific, so why not add it to jsk_recognition_utils/node_scripts? If it is located in jsk_pcl_ros, user need to install jsk_pcl_ros in order to convert 2D polygon array to polygon in pipelines in jsk_perception, for example.

2. Why not change the index as rosparam (dynamic reconfigure is even better) with default value `-1`? I think negative value should be the default and skipped because there is no hint about the number when the node is launched. 

3. Why not change it as the connection-based node? The node [jsk_recognition_utils/node_scripts/pose_array_to_pose.py](https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_recognition_utils/node_scripts/pose_array_to_pose.py) seems a good example.

mmurooka:
Thank you for your advice.
I think all of 1,2,3 advices are reasonable and I created new PR https://github.com/jsk-ros-pkg/jsk_recognition/pull/1826.

Issue #1831, opened 2016-08-07
remove box_array_to_box.py


k-okada:
there are C++ version and python version for same purpose:
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1799/commits/25060c3e517b2a433060762f4afa4e9d2baa7597
https://github.com/jsk-ros-pkg/jsk_recognition/commits/master/jsk_pcl_ros/scripts/box_array_to_box.py

we first find the use case of box_array_to_box.py and consider if it is possible to move to C++ version.

wkentaro:
I sent PR: https://github.com/jsk-ros-pkg/jsk_recognition/pull/1833
But this breaks current API, so it needs to be merged for a major release.
Do you have any idea about the rule about release and breaking API? @k-okada 
For example

|                                    | BugFix | New Node/Nodelets | Break API |
|------------------------------------|--------|-------------------|-----------|
| Major release (ex. 0.3.X -> 1.0.0) | Yes    | Yes               | Yes       |
| Minor release (ex. 0.3.X -> 0.4.0) | Yes    | Yes               | No        |
| Patch release (ex. 0.3.8 -> 0.3.9) | Yes    | No                | No        |

k-okada:
I think the best timing is when we decided to drop hydro support, currently
hrp2 and pr2 is blocker,

--
◉ Kei Okada

wkentaro:
> I think the best timing is when we decided to drop hydro support

I think breaking API and supporting hydro is unrelated.
For example, currently jsk_recognition is already released to hydro, indigo and jade, and
if we break the API the effect is not limited to hydro but indigo and jade.
For this kind of problem, I thought controlling tag version number is most popular solution.
What do you think?

k-okada:
api change and distro upgrade is not related, I just want to push major
 release with some evens, like dropping hydro, start new semester, finished
some event, so users thinks it is reasonable even if they can not
compile/run their program.

--
◉ Kei Okada

wkentaro:
> api change and distro upgrade is not related,

For usual softwares, api change and version number is related, isn't it?
But I understood that for this package they are unrelated.

k-okada:
Oh I mean api change and dropping hydro is not related as you said. and
since we do not create distro specific branch, our software do not have
normal "distro upgrade" process, we'll support newer distro while everyone
(or most of people) using old one. Dropping old distro is only event that
effects to users and that is "distro upgrade" process for us.

--
[...]

PR #1833, opened 2016-08-07
[Major Release] Remove jsk_pcl_ros/box_array_to_box.py

wkentaro:
Close https://github.com/jsk-ros-pkg/jsk_recognition/issues/1831
This change should be released as a major release.

cc @k-okada


k-okada:
why you removed imagesift

--
◉ Kei Okada

wkentaro:
It is already fixed.

k-okada:
since no one claimed that this is used, but not sure if we're ok to remove this .... https://github.com/jsk-ros-pkg/jsk_recognition/pull/1799

k-okada:
it seems we had users for box_array_to_box.py -> https://github.com/jsk-ros-pkg/jsk_recognition/pull/1799#issuecomment-244731229

PR #1839, opened 2016-08-10
Smoothing filter

iory:
Add smoothing filter's, gaussian filter, median filter, bilateral filter.

filter s

PR #1843, opened 2016-08-23
Add nodelet add-image

iory:
add-image

https://github.com/jsk-ros-pkg/jsk_recognition/pull/1839


wkentaro:
The feature of ovelaying **color on mono** already exists. http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/overlay_image_color_on_mono.html
And this nodelet overlays **color on color**, so please give me an example for that.

iory:
Ok, I will show color on color sample. (and changed sample/sample_add_image.launch)

sample

wkentaro:
BTW, what is the use case?

iory:
This is tutorial code for opencv.
https://github.com/opencv/opencv/blob/master/samples/cpp/tutorial_code/core/AddingImages/AddingImages.cpp

For example, output topic of opencv_apps's edge_detector is color image overlayed blue lines.
Using add-image nodelet, we can simply overlay those image to visualize efficiently.
sample_add_image

wkentaro:
I see. The example of visualizing edge detection looks nice. Now I understand the reason why we need this nodelet.

> This is tutorial code for opencv.
https://github.com/opencv/opencv/blob/master/samples/cpp/tutorial_code/core/AddingImages/AddingImages.cpp

If so, why not put this code to opencv_apps? If I remember correctly, @k-okada said the package is created to run OpenCV tutorials/samples with ROS.
And why you changed the name from AddImages to AddImage?

k-okada:
I think it is important to define the list of OpenCV tutorials/samples to
be created as ros node, otherwise we have to ceate ros sample code forever,

In https://github.com/ros-perception/opencv_apps/blob/
indigo/CMakeLists.txt#L103, I created list of OpenCV code from
https://github.com/opencv/opencv/tree/2.4/samples/cpp/tutorial_code, but it
becomes old.

I have updated the list of opencv tutorial code, so if you PR  (
https://github.com/ros-perception/opencv_apps/pull/25), I'm very hapy to
merge it.


--
◉ Kei Okada

iory:
OK

PR #1848, opened 2016-08-25
Retry at most three times point_indices_to_mask_image.test

wkentaro:
To fix error sometimes on Travis.

### DIFF
--- a/jsk_pcl_ros_utils/test/point_indices_to_mask_image.test
+++ b/jsk_pcl_ros_utils/test/point_indices_to_mask_image.test
@@ -2,7 +2,8 @@
 
   <test test-name="test_point_indices_to_mask_image"
         name="test_point_indices_to_mask_image"
-        pkg="jsk_pcl_ros_utils" type="test_point_indices_to_mask_image.py">
+        pkg="jsk_pcl_ros_utils" type="test_point_indices_to_mask_image.py"
+        retry="3">
     <remap from="~input/mask1" to="point_indices_to_mask_image/output" />
     <remap from="~input/mask2" to="point_indices_to_mask_image_static/output" />
   </test>


k-okada:
why it sometimes fail?

--
◉ Kei Okada

wkentaro:
Sometimes the topic message does not come and the test reaches the timeout.
It happens only on Travis/Jenkins, and I don't know other ways to fix this than retrying.

k-okada:
may be because of lack of cpu resources, try test code under high load average or cheapest AWS machine.
ONe solution may running node with `use_sim_time` with slow `/clock` speed, and other may be increase queue size for the subscriber?
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/src/point_indices_to_mask_image_nodelet.cpp#L60-L66

wkentaro:
You mean retrying is not good solution?

k-okada:
Yes, I think that should be used when you know that process is designed to stochastically fail.
In this case, you created the node as to work 100% reliability, so this means there might be something wrong.
Usually if it is working on you machine, but failing in travis/jenkins usually means the node works wrongly if the CPU/memory resource is limited.
So it is better to remember that there might be bug in point_indices_to_mask_image_nodelet.cpp

One question is in the following code, queue size of input and input/image is 1, but that of message filter is 100. Is this ok?
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/src/point_indices_to_mask_image_nodelet.cpp#L60-L66

wkentaro:
> queue size of input and input/image is 1

is for messages for processing in callback.

> that of message filter is 100

is for sychronization of messages.

So I think it is ok.

---

> Yes, I think that should be used when you know that process is designed to stochastically fail.
In this case, you created the node as to work 100% reliability, so this means there might be something wrong.
Usually if it is working on you machine, but failing in travis/jenkins usually means the node works wrongly if the CPU/memory resource is limited.
So it is better to remember that there might be bug in point_indices_to_mask_image_nodelet.cpp

Hmm, but currently we use retry=3 for almost all test files...
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/test/bounding_box_array_to_bounding_box.test
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/test/cluster_point_indices_to_point_indices.test
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/test/point_indices_to_cluster_point_indices.test
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/test/pointcloud_to_mask_image.test
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros_utils/test/pointcloud_to_point_indices.test
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/test/add_mask_image.test
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/test/extract_image_channel.test

k-okada:
>
> So I think it is ok.

if so, message filter has 100 x 2 queues and copy the data from
subscriber, instead of just to have an array of pointer. that is not
efficient....
anyway i just look at the test code and it seems very simple and not
sure why test fails, so only possibility is to drop incoming data by
setting queue size 1. May be there might be other problems.

> Hmm, but currently we use retry=3 for almost all test files...

Yes, and that's may hiding potential bugs. I do not say you have to
fix that, just to remind that adding "retry" did not "solve" problem,
the bug may appear in future, sometimes just before the deadline or
during important demos....

PR #1852, opened 2016-08-30
Add missing build_depend on jsk_data

wkentaro:
This is necessary to run install script on CMakeLists.txt.


k-okada:
which install script?

wkentaro:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/CMakeLists.txt#L426

PR #1828, opened 2016-08-06
Fix configuration of jsk_pcl_ros

wkentaro:

PR #1857, opened 2016-09-06
Fix configuration of jsk_pcl_ros

k-okada:
rewrite #1828

1. just change the order in find_package
2. do not have to add genmsg to build_depend
3. do not add 'missing components' in find package (https://github.com/jsk-ros-pkg/jsk_recognition/pull/1828/commits/c54d279c3c689d649e0e5ce6e6423911639686f0)

by the wey, why did you add these missing components? did `catkin_lint` said so?

- Missing installation of executables
- Fix missing dependency declaration of jsk_pcl_ros
- Fix order of components in find_package of jsk_pcl_ros

### DIFF
--- a/jsk_pcl_ros/CMakeLists.txt
+++ b/jsk_pcl_ros/CMakeLists.txt
@@ -2,26 +2,34 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(jsk_pcl_ros)
 
-# Load catkin and all dependencies required for this package
-# TODO: remove all from COMPONENTS that are not catkin packages.
-set(PCL_MSGS pcl_msgs) ## hydro and later
-
 find_package(catkin REQUIRED COMPONENTS
-  jsk_recognition_utils jsk_pcl_ros_utils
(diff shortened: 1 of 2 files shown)


wkentaro:
> by the wey, why did you add these missing components? did catkin_lint said so?

Yes. All dependencies that are listed as `build_depend` in `package.xml` should be listed in `find_package`, right?

k-okada:
I think catkin_lint assuming that all packages in build_depends has
different <PACKAGE>_INLCUDE_DIRS/LIBRARIES, that we need to add them to
find_package,
If we can compile current code with current find_package, i think we can
remove packages from build_depend, instead of adding packages find_package.

--
◉ Kei Okada

wkentaro:
Ok, so you'd like to remove the declarations in package.xml.

k-okada:
yes, always try to reduce size of sources
13. ``Perfection (in design) is achieved not when there is nothing more to add, but rather when there is nothing more to take away.''  http://cruel.org/freeware/cathedral.html#6

PR #1855, opened 2016-09-05
[jsk_pcl_ros_utils]  unwrap polygon array with likelihood

furushchev:
https://github.com/furushchev/jsk_recognition/blob/42c0018aa8dab7ffe0d164c7ad92f64e16544849/doc/jsk_pcl_ros_utils/nodes/polygon_array_unwrapper.rst

PR #1858, opened 2016-09-06
[jsk_pcl_ros_utils] unwrap polygon array with likelihood

k-okada:
fixed versio of #1855


k-okada:
Closes #1855

furushchev:
@k-okada Thank you very much for heliping!

Issue #1854, opened 2016-08-30
release 0.4.0


k-okada:
0.3.22
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1849
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1835
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1829
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1828

~~0.4.0~~ 1.0.0
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1833
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1827
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1820
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1791
- [x] https://github.com/jsk-ros-pkg/jsk_recognition/pull/1726

k-okada:
0.3.22 done  https://github.com/ros/rosdistro/pull/12681

wkentaro:
It's done. https://github.com/jsk-ros-pkg/jsk_recognition/releases/tag/1.0.0

PR #1861, opened 2016-09-07
Revert "Remove dependency on jsk_perception for separated build"

k-okada:
Reverts jsk-ros-pkg/jsk_recognition#1820

But is that ok to merge this before releasing 0.3.22?
It seems better to release this at 0.4.0: #1854


wkentaro:
LGTM

PR #1263, opened 2015-10-19
Adaptive object tracking

iKrishneel:
Nodelet for updating template during tracking using the Particle Filters. 
-- the nodelet for selecting the target object for tracking will be will be added soon with the descriptions

PR #1537, opened 2016-02-17
[jsk_perception] Add cmt nodelet2

garaemon:
Depends on https://github.com/jsk-ros-pkg/jsk_3rdparty/pull/56 and https://github.com/aginika/libCMT/pull/1

PR #1868, opened 2016-09-12
Adaptive object tracking

k-okada:
rewrited version of #1263


k-okada:
I could not take into account https://github.com/jsk-ros-pkg/jsk_recognition/pull/1263#issuecomment-149437955 comment, @iKrishneel 
it is also better to compare this method with another one, like https://github.com/jsk-ros-pkg/jsk_recognition/pull/1537 
One big benefit of having jsk_recognition repository is you can easily compare your work with others, that will strength your paper.

PR #1875, opened 2016-09-15
fix TargetAdaptiveTrackingConfig file name (this breaks make install)

k-okada:


k-okada:
Need to check why we can't catch this issue

wkentaro:
Possibly because:

1. Currently install test is disabled for jsk_recognition packages.
2. `generate_dynamic_reconfigure_options()` macro does not check file existence (?)

wkentaro:
I found `generate_dynamic_reconfigure_options()` checks file existence, but the Header file is no generated based on the basename of the `.cfg` file,
but the third argument for `gen.generate()` function.
https://github.com/k-okada/jsk_recognition/blob/47e47d3a90c41177407f6d4477f81967729957fc/jsk_pcl_ros/cfg/TargetAdaptiveTracking.cfg#L31
If it was `"TargetAdaptiveTrackingConfig"`, it works without changing the name of file. (of course changing filename is better in this case.

PR #1877, opened 2016-09-16
Remove no need CATKIN_TOOLS_BUILD_OPTIONS declaration

wkentaro:
~~For https://github.com/jsk-ros-pkg/jsk_recognition/pull/1875#issuecomment-247397627~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
     - NOT_TEST_INSTALL=true
     - USE_JENKINS=true
   matrix:
-    - ROS_DISTRO=hydro  USE_DEB=false CATKIN_TOOLS_BUILD_OPTIONS="-iv --summarize --no-status --no-color"
+    - ROS_DISTRO=hydro  USE_DEB=false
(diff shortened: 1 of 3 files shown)


wkentaro:
Please merge this.

k-okada:
Does removing 'CATKIN_TOOLS_BUILD_OPTIONS" from hydro test relate to 
```
I found generate_dynamic_reconfigure_options() checks file existence, but the Header file is no generated based on the basename of the .cfg file,
but the third argument for gen.generate() function.
https://github.com/k-okada/jsk_recognition/blob/47e47d3a90c41177407f6d4477f81967729957fc/jsk_pcl_ros/cfg/TargetAdaptiveTracking.cfg#L31
If it was "TargetAdaptiveTrackingConfig", it works without changing the name of file. (of course changing filename is better in this case.
```
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1875#issuecomment-247594262

wkentaro:
That was wrong comment, sorry.

PR #1899, opened 2016-10-10
Add .gitattributes file for GitHub statistics

wkentaro:
https://github.com/github/linguist#using-gitattributes

PR #1900, opened 2016-10-10
Highlight syntax of *.test on GitHub

wkentaro:
screen shot 2016-10-11 at 7 56 25 am

screen shot 2016-10-11 at 7 56 10 am

### DIFF
--- a/imagesift/test/test_imagesift.test
+++ b/imagesift/test/test_imagesift.test
@@ -1,3 +1,4 @@
+<!-- vim: set ft=xml: -->
 <launch>
 
   <include file="$(find imagesift)/sample/imagesift_sample.launch">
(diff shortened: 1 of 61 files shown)


wkentaro:
Please merge this.

k-okada:
humm, why https://github.com/jsk-ros-pkg/jsk_recognition/pull/1899 did not take care of this...

wkentaro:
It is described here, but there is no reason .https://github.com/github/linguist#using-gitattributes

wkentaro:
It seems that this solution is not good because it is not automated and we should **always** add the modeline.
So I close this at this time.

k-okada:
good decision ;)

--
◉ Kei Okada


2016-10-15 5:54 GMT+09:00 Kentaro Wada <notifications@github.com>:

PR #1910, opened 2016-10-20
[jsk_pcl_ros] Add warning output to screenpoint nodelet

mmurooka:


jsk_pcl_ros/src/pointcloud_screenpoint_nodelet.cpp
```diff
@@ -132,10 +132,14 @@ void jsk_pcl_ros::PointcloudScreenpoint::onInit()
 
 bool jsk_pcl_ros::PointcloudScreenpoint::checkpoint (pcl::PointCloud< pcl::PointXYZ > &in_pts, int x, int y,
                                                      float &resx, float &resy, float &resz)  {
-  if ((x < 0) || (y < 0) || (x >= in_pts.width) || (y >= in_pts.height)) return false;
+  if ((x < 0) || (y < 0) || (x >= in_pts.width) || (y >= in_pts.height)) {
+    JSK_ROS_WARN_STREAM("Requested point is out of image size.  point: ("
+                        << x << "," << y << ")" << "  size: (" << in_pts.width << "," << in_pts.height << ")");
```
wkentaro: Why not use `JSK_ROS_WARN("... point: (%d, %d) ..")`?

wkentaro: Using `(%d, %d)` is more readable don't you think?

mmurooka: How do you switch to use stream and printf? I'd like to know general criteria.
Maybe I'll do same mistake in the next time, and I'd like to reduce chance to modify PR.

wkentaro: Personally I usually use printf, and I use stream in the case:
```
ROS_INFO_STREAM("Number of clusters: " << cluster_indices_msg.clusters.size());
```
or when displaying rosmsg
screen shot 2016-10-21 at 11 43 54 pm

wkentaro: > Maybe I'll do same mistake in the next time, and I'd like to reduce chance to modify PR.

Anyway, you can ignore my comment, because this is just a comment.

mmurooka: Updated.

> ROS_INFO_STREAM("Number of clusters: " << cluster_indices_msg.clusters.size());

What is the feature of this case?

wkentaro: Looks better to me.

mmurooka:
Apply @wkentaro 's comment, and travis test passed.

PR #1925, opened 2016-10-25
[jsk_pcl_ros] fix typo in jsk_pcl_nodelets.xml

knorth55:
we can compile jsk_pcl_ros but we cannot load nodelets written after this typo

### DIFF
--- a/jsk_pcl_ros/jsk_pcl_nodelets.xml
+++ b/jsk_pcl_ros/jsk_pcl_nodelets.xml
@@ -631,6 +631,7 @@
   <class name="jsk_pcl/CollisionDetector"
          type="jsk_pcl_ros::CollisionDetector"
          base_class_type="nodelet::Nodelet">
+  </class>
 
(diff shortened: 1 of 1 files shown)


k-okada:
nice catch, but why we can't find this on travis test?

--
◉ Kei Okada


2016-10-25 20:45 GMT+09:00 Shingo Kitagawa <notifications@github.com>:

knorth55:
if we have test of `TargetAdaptiveTracking`, we can get to know easily to this bug.
or we can have xml syntax test for this file, too.

anyway, catkin doesn't say any error on this kind of typo.

k-okada:
I think better solution is to create test code that runs TargetAdaptiveTracking, this is similar strategy that we discussied on your thesis yesterday, instead of checking every step, we could check entire system if it works or not

But my questions is; did test code using jsk_pcl_nodelets.xml runs without any failure?

knorth55:
that is because ros/pluginlib does not check syntax of xml and parse it from top.
after https://github.com/ros/pluginlib/pull/4, it will throw error

cc @wkentaro 

https://github.com/ros/pluginlib

k-okada:
I see thanks,

PR #1927, opened 2016-10-25
[jsk_pcl_ros] add xmllint syntax check

knorth55:
related to #1925


wkentaro:
Checking syntax **after** the build is not undesirable (because user needs to wait for that on CI),
roslint is a tool for that https://github.com/ros/roslint with
```
catkin roslint --this --no-deps -iv
```

knorth55:
should be in `lint.bash` #1928

k-okada:
thanks, once http://build.ros.org/job/Ibin_uT64__pluginlib__ubuntu_trusty_amd64__binary/ is done, we can check within existing test code, I think

PR #1929, opened 2016-10-25
[jsk_pcl_ros] fix unparsable jsk_pcl_nodelets.xml

knorth55:
1. multiple <library> tags in one xml file can't be used.
2. separate pluginlib xml for each library files.
3. pluginlib xml should be splitted with package name
   (nodelet, moveit_ros_perception).


k-okada:
does this means that these nodelets are not used so far?

knorth55:
this PR means
- correct xml syntax (one`library` tag for each file, and `library` tag should be top of xml tags)

in order to correct xml syntax, split `jsk_pcl_nodelets.xml` to three files and arrange them in `plugins` dir.

`libjsk_pcl_ros` and `libjsk_pcl_ros_utils` are nodelet so that we put them in `plugins/nodelet` dir, but we cannot load `libjsk_pcl_ros_moveit` nodelet.
We thought it is not nodelet so that we put it in `plugins/moveit_ros_perception`

anyway, we can use all nodelets as before, but just fix xml syntax and arrange them.

cc. @wkentaro

k-okada:
is there any chance to fix paring library for this file? I think having same ta for one file is ok for XML

wkentaro:
It seems that below syntax is bad for xml
```xml
<library>
</library>
<library>
</library>
```
and it should be:
```xml
<xml>
  <library>
  </library>
  <library>
  </library>
</xml>
```
but this is not supported by pluginlib.

k-okada:
what I understand is;

- nodelet.xml can have multiple libraries if it used with pluginlib
- `rospack plugins --attrib=plugin nodelet` uses xmllib and that did not read file with multiple root

knorth55:
in previous xml style, we cannot find `jsk_pcl` nodelets by `rospack plugins`, that's why we thought nodelets.xml was not written in correct xml syntax

k-okada:
please please please do not change directory layout just because you do not like that.....
http://build.ros.org/job/Ibin_uT64__jsk_perception__ubuntu_trusty_amd64__binary/60/

knorth55:
@k-okada sorry for ourt carelessness
@wkentaro :innocent:

PR #1931, opened 2016-10-26
[jsk_perception] Improve the generality of the fisheye rectification process

tongtybj:
1. Use the encoding variable of sensor_msgs:image while subscribing,and same depth for undistorted image. The previous fixed sensor_msgs::image_encodings(BGR8) is not compatible with mono-scale image
2. Add configurable for the variable k, which was fixed for prosillica.


k-okada:
> gimbal func for fisheye

why you do not use tf for this.

k-okada:
> why you do not use tf for this.

it may be not the direction of the camera, but the optical axis for rectified image, any way please add/update document for this node.

PR #1935, opened 2016-10-27
[checker_board_detector] check if image already rectifierd

YuOhara:
This PR is to prevent rectified image been rectified again.
Check if the image already rectified or not by topic having "image_rect" in its name.

### DIFF
--- a/checkerboard_detector/src/checkerboard_detector.cpp
+++ b/checkerboard_detector/src/checkerboard_detector.cpp
@@ -91,6 +91,7 @@ public:
     int message_throttle_counter_;
     string frame_id; // tf frame id
     bool invert_color;
+    bool rectify_flug1, rectify_flug2;
     int display, verbose, maxboard;
     vector<CHECKERBOARD> vcheckers; // grid points for every checkerboard
     vector< string > vstrtypes; // type names for every grid point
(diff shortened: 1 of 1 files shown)


k-okada:
i think we should not depend on topic name, how about
https://github.com/ros-perception/image_pipeline/blob/indigo/image_proc/src/nodelets/rectify.cpp#L117-L122

YuOhara:
> i think we should not depend on topic name, how about

Yes, i think so.
But there seems no info about if image is rectified or not.
Because image_raw and image_rect uses the same camera_info

```
rosmsg show sensor_msgs/Image 
std_msgs/Header header
  uint32 seq
  time stamp
  string frame_id
uint32 height
uint32 width
string encoding
uint8 is_bigendian
uint32 step
uint8[] data
```

Also, I can't understand why
https://github.com/ros-perception/image_pipeline/blob/indigo/image_proc/src/nodelets/rectify.cpp#L117-L122
needed.

k-okada:
> https://github.com/ros-perception/image_pipeline/blob/indigo/image_proc/src/nodelets/rectify.cpp#L117-L122

I see, that's checking if camera is calibrated or not, and that's different from rectified or not.

YuOhara:
Many users use relay tools (like that in jsk_topic_tools) that changes topic names, so this pr seems to be confusing.

PR #1938, opened 2016-10-27
option for not rectify image

YuOhara:
seems better than #1935 ?

### DIFF
--- a/checkerboard_detector/src/checkerboard_detector.cpp
+++ b/checkerboard_detector/src/checkerboard_detector.cpp
@@ -99,6 +99,7 @@ public:
     int dimx, dimy;
     bool use_P;
     double fRectSize[2];
+    bool rectify_flug_;
 
     //////////////////////////////////////////////////////////////////////////////
     // Constructor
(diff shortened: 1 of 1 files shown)


YuOhara:
it has no efect to existing code because param is set true as default

k-okada:
I think this is very difficult if you apply rectification process twice. so seems https://github.com/jsk-ros-pkg/jsk_recognition/pull/1935 is better, 

how about measure distance between actual checker corners and they are in model. if the camera is well calibrated and rectified, it should be 0, and if that error is larger than some value, then warn it.

YuOhara:
> how about measure distance between actual checker corners and they are in model. if the camera is well calibrated and rectified, it should be 0, and if that error is larger than some value, then warn it.

Sounds good if it can be done with few additional code and little execution time.
I will try it tomorrow.

YuOhara:
> I think this is very difficult if you apply rectification process twice

point_pose_extractor uses zero distortion model
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/src/point_pose_extractor.cpp#L394

k-okada:
- I think this is very difficult if you apply rectification process twice
+ I think this is very difficult to realize if you apply rectification process twice

PR #1939, opened 2016-10-28
[WIP] [checkerboard_detector] warn if error is big

YuOhara:
do not merge yet.
I will check parameters tomorrow with other calibrated cameras (I checked only on my astra camera) and 5X4 checkerboard.
any comments are welcome.

With this PR, we can warn dual-rectification of image indicated in #1935 #1937.
However, the error often occurs because of moving cameras or checker board is not in planer place.

### DIFF
--- a/checkerboard_detector/src/checkerboard_detector.cpp
+++ b/checkerboard_detector/src/checkerboard_detector.cpp
@@ -541,7 +541,7 @@ public:
                                          cv::Size(size,size), cv::Size(-1,-1),
                                          cv::TermCriteria(CV_TERMCRIT_ITER, 50, 1e-2));
                     }
-                    objpose.pose = FindTransformation(cb.corners, cb.grid3d, cb.tlocaltrans, model);
+                    objpose.pose = FindTransformation(cb.corners, cb.grid3d, cb.tlocaltrans, model, size);
                 }
 
 #pragma omp critical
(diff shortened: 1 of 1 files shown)


k-okada:
> checker board is not in planer place.

please do not use such checker board, please do not use one just because it is already there, make an effort for better solution if your program depends on this.
I have ordered plastic board printed checker board (@mmurooka, @yuki-asano ) and check it that works for you

YuOhara:
i just wrote code corresponding to comment in #1937
> how about measure distance between actual checker corners and they are in model. if the camera is 
> well calibrated and rectified, it should be 0, and if that error is larger than some value, then warn it.

And i found that this code can check not only dual-rectification of images, but also the flatness of the checker board. 
I hope checkerboard which is not so flat will be detected by this PR, and removed.

YuOhara:
I tested with checkerboard which i printed just for testing this PR.
I do not mean that i used checkerboards which are used in demo.

yuki-asano:
> I have ordered plastic board printed checker board (@mmurooka, @yuki-asano ) and check it that works for you

The checker board is working with this program, and is good for me.
https://github.com/jsk-enshu/robot-programming/blob/master/dxl_armed_turtlebot/euslisp/display-checkerboard.l

mmurooka:
> The checker board is working with this program, and is good for me.
https://github.com/jsk-enshu/robot-programming/blob/master/dxl_armed_turtlebot/euslisp/display-checkerboard.l

@k-okada, I get one checkerboard from @yuki-asano and this seems to be useful. I'd like to get different size of board. Could you tell where you get this?

PR #1943, opened 2016-10-30
Fix for kinetic build

k-okada:
-    .travis.yml: run kinetic build, without test
-    use std::isnan instead of isnan, knetic compiler requires this
-    CMakeFiles.txt : add c++11 option if possible,    http://answers.ros.org/question/152276/is-there-a-way-to-enable-c11-support-for-catkin-packages/
-    remove depends to driver_base


k-okada:
currently https://github.com/pr2-gbp/robot_self_filter-gbp is only blocking package

PR #1964, opened 2016-11-18
[Wip] Kinfu publish tsdf

YuOhara:
fyi @mmurooka 
the picture below shows empty volume 
screenshot from 2016-11-18 21 09 57

### DIFF
--- a/doc/jsk_pcl_ros/nodes/kinfu.md
+++ b/doc/jsk_pcl_ros/nodes/kinfu.md
@@ -19,3 +19,8 @@ Publishes camera pose using pcl/KinfuLS
 * `~output/cloud` (`sensor_msgs/PointCloud2`)
 
   Registered scene pointcloud
+
+## Advertising Services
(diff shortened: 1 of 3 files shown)


YuOhara:
updated, thanks.

k-okada:
> 	[jsk_pcl_ros/kinfu] publish tsdf if service is called

1, why does tsdf stand for?
2. why you use service to publish, is there any problem if it publishes only when subscribed

YuOhara:
Tsdf stands for `Truncated Signed Distance Function.`
I used service because this process takes too much time. (somehow because I just implemented it with CPU, not GPU).

I also thinks that service for publishing is not good. Because service-caller-node cannot get published msg.
Add msg for Tsdf and include it in service responce seems better.

YuOhara:
i will update pr.

wkentaro:
> Add msg for Tsdf and include it in service responce seems better.

+1

PR #1969, opened 2016-11-22
[jsk_pcl_ros/pointcloud_dataserver] remove adding unneeded cloud

YuOhara:
Cloud msgs grows periodly because the same cloud is added when timer-callback is driven.
Therfore, moved to Initialization of  cloud msg.

This may be bug-fix. changed in https://github.com/jsk-ros-pkg/jsk_recognition/pull/1926
@knorth55  please check this.


knorth55:
thank you, it's my mistake. 
+1

PR #1970, opened 2016-11-22
Kinfu get tsdf

YuOhara:
updated version of #1964 
* the service returns raw tsdf values
* if subscriber exists, publish cloud
also updated doc


YuOhara:
@mmurooka 
I found that using nodelet prevents message-size-limit-error.
So please use nodelet to resize or crop cloud.

k-okada:
I do not think publishing topic within service call is not good manner
basically ros node is designed to create data stream, service call is to control the behaviors, not to get information. Imagine if you can calculate if you can compute tsdf within a framerate, you'll not design same way,.

PR #1975, opened 2016-11-29
[jsk_recogintion_msgs][Rect.msg] add value/label properties for Rect.msg

furushchev:
like `BoundingBox.msg`.

### DIFF
--- a/jsk_recognition_msgs/msg/Rect.msg
+++ b/jsk_recognition_msgs/msg/Rect.msg
@@ -2,3 +2,6 @@ int32 x
 int32 y
 int32 width
 int32 height
+# You can use this field to hold value such as likelihood
+float32 value
+uint32 label


wkentaro:
IMO, adding label is not necessary, because we can synchronize with other topic or use rosparam to get the label.


If you really like this feature, I suggest adding `Label.msg` and generate `RectLabeled.msg` like below:

```
# jsk_recognition_msgs/Label.msg
int32 label
string label_name
float32 likelihood
```

```
# jsk_recognition_msgs/RectLabeled.msg
jsk_recognition_msgs/Label label
jsk_recognition_msgs/Rect rect
```

furushchev:
@wkentaro 
- Can we identify classes of each rectangle in one image without labeling if we synchronize with other topics and rosparam? How can we do it?
- Is it ok to add `Label.msg` and add label properties to `Rect.msg`, `BoundingBox.msg` and so on. (It may break existing label property)

wkentaro:
> Can we identify classes of each rectangle in one image without labeling if we synchronize with other topics and rosparam? How can we do it?

If we synchronize with other topics or rosparam, we can identify the classes, no?
Anyway, without knowing your use case, I think I will be missing something you're proposing.

> Is it ok to add Label.msg and add label properties to Rect.msg, BoundingBox.msg and so on. (It may break existing label property)

I don't think it is ok, so we need to create another message like RectLabeled.msg.

wkentaro:
As an **example**, if you'd like to publish and visualize object detection result,
I propose a system like below:

- `object_detections` publishes DetectedObjectArray.msg.
- `draw_object_detections` synchronize Image.msg and DetectedObjectArray.msg and visualize.

```
# DetectedObject.msg
string label  # object name or id
float64 confidence  # object detection confidence
jsk_recognition_msgs/Rect rect  # object location

# DetectedObjectArray.msg
Header header
string[] labels  # object all names or ids
jsk_recognition_msgs/DetectedObject[] objects
```

furushchev:
@wkentaro thank you for explanation. I partially agree that it's good to create message files specified to applications in ros  ecosystem. But I'd like not to create another message file because we must support in all programs which we'd like to use them.
Adding properties in existing message files does not increase overhead of data if they are not used, and compatible to all programs without any fix.

wkentaro:
## Response

> But I'd like not to create another message file because we must support in all programs which we'd like to use them.
Adding properties in existing message files does not increase overhead of data if they are not used, and compatible to all programs without any fix.

It is true as you did in draw_rect_array.py
- https://github.com/jsk-ros-pkg/jsk_recognition/pull/1976/files#diff-4b8ae7d9cfcd28b6e36c9b2f736c2172R18

but it does not use the label field and the result is a bit useless, I think.
If you are changing draw_rect_array.py, there are few differences between adding new subscription of `RectLabeledArray.msg` and adding `if rect_msg.label != 0: ...`.


## My idea

The problem of adding new field is 
1.  It is very difficult if we add once. (so we need discussion before adding it like in this PR)
1.  There are no value like `None` or `NULL` (**NULL** value) for the ROS message, so we cannot identify `0` for `uint8 label` is the default or not specified. (currently I'm not sure whether adding `string label` is reasonable because `""` represents the **NULL** value)
1. Rect is Rect, and it is not reasonable (for me) with label or value, as most messages in [geometry_msgs](https://github.com/ros/common_msgs/blob/jade-devel/geometry_msgs/msg/Pose.msg).

wkentaro:
@jsk-ros-pkg/developers  Any other ideas?

mmurooka:
+1 for making RectLabeled.msg and not changing Rect.msg

furushchev:
@mmurooka I'd like to know why :)

mmurooka:
It doesn't effect current working program. (if Rect.msg is used across other PCs and only one PC is updated, md5sum problem occurs.) If this is very important feature, it is OK, but I think this case is not.
If you visualize rect with label, maybe you'd like to colorize each rect according to label value, and in that case, you need to add source code to visualize program in both cases of changing Rect.msg and adding RectLabeled.msg. If you are not interested in label value when visualizing, you can publish Rect.msg and RectLabeled.msg and use Rect.msg for visualizing and RectLabeled.msg for processing.

But, I'm not heavy user of jsk_recognition, so please consider my +1 as around +0.2.

furushchev:
Thank you for clarification. I'll then make another messages. (Though BoundingBox.msg has still label and value).

wkentaro:
> Though BoundingBox.msg has still label and value

Yeah, I'd like to remove it. (the header in BoundingBox.msg also because BoundingBoxArray.msg has header for each boxes, and that is unusual)

furushchev:
@wkentaro In my opinion, it's better that each instance has a header because we cannot use message without header after task performance (it is bad for logging). (And even if each instance has a header, it does not have any overhead if we let them empty when we publish as array.

wkentaro:
> In my opinion, it's better that each instance has a header because we cannot use message without header after task performance (it is bad for logging)

Why not we create `BoundingBoxStamped.msg`? (as the willow do in geometry_msgs)

> And even if each instance has a header, it does not have any overhead if we let them empty when we publish as array.

I mean it is confusing for both publishers and subscribers. I think clarity is one of the most important things when we communicate with messages. (like Can I remain it as empty? Should I fill with a same headers? Can I assume they are all same? ... etc.)
If the developer of publisher (like garaemon in drc) and subscriber (like mmurooka in drc) is different, it causes trouble I think.

@mmurooka @YuOhara @k-okada What do you think?

PR #1978, opened 2016-12-04
[jsk_pcl/ICP] change max param of icp-cfg(debug)

YuOhara:
seems just a mistake.


YuOhara:
not able to select NDT with this change

k-okada:
- not able to select NDT with this change
+ not able to select NDT without this change
?

YuOhara:
sorry for my mistake. the latter is true.
> not able to select NDT without this change

PR #1981, opened 2016-12-05
[jsk_recognition_msgs] add Label, LabeledRect, LabeledRectArray

furushchev:
ref: #1975

### DIFF
--- a/jsk_recognition_msgs/msg/Label.msg
+++ b/jsk_recognition_msgs/msg/Label.msg
@@ -1,0 +1,3 @@
+int32 index
+string name
+float32 likelihood
--- a/jsk_recognition_msgs/msg/LabeledRect.msg
+++ b/jsk_recognition_msgs/msg/LabeledRect.msg
@@ -1,0 +1,2 @@
+Rect rect
+Label label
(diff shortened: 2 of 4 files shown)


jsk_recognition_msgs/msg/Label.msg
```diff
@@ -0,0 +1,3 @@
+int32 index
```
wkentaro: I think `label` or `id` is more appropriate name, because `index` is usually non-negative and it must be with a list of values.

jsk_recognition_msgs/msg/LabeledRect.msg
```diff
@@ -0,0 +1,2 @@
+Rect rect
+Label label
```
wkentaro: Why not place `label` at the first line, like `Header`?
```
Label label
Rect rect
```

k-okada:
@furushchev any updates?

furushchev:
@k-okada This pull request is actually overriden by #1982. And that is on my queue. Please wait a moment...

k-okada:
I'm waiting ... @furushchev

furushchev:
I found that we can use `jsk_recognition_msgs/RectArray` and `jsk_recognition_msgs/ClassificationResult` as a result of 2d object detection instead of introducing new messages. See #2205 for more detail.

PR #1982, opened 2016-12-05
[jsk_perception] add draw_labeled_rect

furushchev:
for visualization


jsk_perception/src/draw_labeled_rect.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  void DrawLabeledRect::unsubscribe()
+  {
+    sub_image_.unsubscribe();
+    sub_rects_.unsubscribe();
+  }
+
+  void DrawLabeledRect::getColorString(const jsk_recognition_msgs::Label &label, cv::Scalar& color)
```
wkentaro: What do you mean with `ColorString`?

jsk_perception/src/draw_labeled_rect.cpp
```diff
@@ -0,0 +1,147 @@
...
+  {
+    static const float colors[6][3] = { {1,0,1}, {0,0,1},{0,1,1},{0,1,0},{1,1,0},{1,0,0} };
```
wkentaro: Why not use:

- https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_recognition_utils/src/rgb_colors.cpp

jsk_perception/src/draw_labeled_rect.cpp
```diff
@@ -0,0 +1,147 @@
...
+    static const double font_size = 1.5;
+    static const int font_thickness = 5;
+
+    cv::Scalar color;
+    getColorString(msg.label, color);
+
+    cv::Rect rect(msg.rect.x, msg.rect.y, msg.rect.width, msg.rect.height);
+
+    cv::rectangle(mat, rect, color, 2, CV_AA);
+    if (!msg.label.name.empty()) {
```
wkentaro: Why not:

```
if (msg.label.name.empty()) {
  return;
}
```

jsk_perception/src/draw_labeled_rect.cpp
```diff
@@ -0,0 +1,147 @@
+/*********************************************************************
```
wkentaro: Why not name it as `draw_rects.cpp` or `draw_rect_array.cpp` with subscription of `~input/rect_array` and `~input/labeled_rect_array`, with adjustments of code for both `RectArray` and `LabeledRectArray`?

wkentaro: @k-okada I think it is possible with some change about the published topics as recognition results.
### VGG for classification: 

- Recognition: http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/vgg16_object_recognition.html
- Visualization: http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/draw_classification_result.html

## FasterRCNN for detection

- Recognition & Visualization: http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/fast_rcnn.html

k-okada:
or can we change darknet node to publish same topics as existing node?
do we already discuss this topic at
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1975 ?

--
◉ Kei Okada


2017-01-21 9:53 GMT+09:00 Kentaro Wada <notifications@github.com>:

wkentaro:
> or can we...

current nodes won't publish information much enough to visualize as a
detection result. That's why the labeledrect is introduced.

And yes, that's what is discussed on the PR.


2017年1月21日(土) 10:20 Kei Okada <notifications@github.com>:

k-okada:
@wkentaro so can you update fast_rcnn.cpy to publish labled rect ? so that we can visualize the result using draw_labeled_rect?
@furushchev can you create classification example using darknet which publish jsk_recognition_msgs/ClassificationResult ? so that we can visualize darknet node output using draw_classification_result.py

@wkentaro @furushchev  detection and classification result visualization tools are `draw_labeled_rect` and `draw_classification_result.py`, both uses 'draw_'+'topic name' so that's ok. but seems like one uses `draw_` + `general data type` and one uses `draw_` + `data semantic meaning`. Related disussion -> https://discourse.ros.org/t/suggestions-for-std-srvs/1079/15?u=k-okada

wkentaro:
> @wkentaro so can you update fast_rcnn.cpy to publish labled rect ? so that we can visualize the result using draw_labeled_rect?

Yeah that is possible.

> @furushchev can you create classification example using darknet which publish jsk_recognition_msgs/ClassificationResult ? so that we can visualize darknet node output using draw_classification_result.py

Actually that seems not possible, because jsk_recognition_msgs/ClassificationResult does not have information about location, ROI.
Because of the difference between classification and detection.

> @wkentaro @furushchev detection and classification result visualization tools are draw_labeled_rect and draw_classification_result.py, both uses 'draw_'+'topic name' so that's ok. but seems like one uses draw_ + general data type and one uses draw_ + data semantic meaning. Related disussion -> https://discourse.ros.org/t/suggestions-for-std-srvs/1079/15?u=k-okada

That should be always considered for the selection of good msg and topic names.
In this case, we can use `geometry_msgs/PolygonStamped.msg` instead of `jsk_recognition_msgs/Rect.msg`, but former is usually used for 3D and that's why furushchev decided to create `LabeledRectArray.msg` instead of `LabeledPolygonArray.msg`, so that the msg name and structure prevent the misuse of the published topic, I think.

jsk_perception/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ find_package(catkin REQUIRED COMPONENTS
   angles
   cmake_modules
   cv_bridge
+  darknet
```
furushchev: I tried to compile with darknet from deb, but it seems to be missing compile settings from my another PR ( https://github.com/jsk-ros-pkg/jsk_3rdparty/pull/83 ) and I could not compile at all.
I'm now continuing improving this PR with `jsk_3rdparty` version of `darknet` package.

jsk_perception/include/jsk_perception/darknet_object_detection.h
```diff
@@ -0,0 +1,145 @@
...
+
+namespace jsk_perception
+{
+
+class AverageArray {
+  float ** arr;
+  bool valid;
+  int n, e, c;
+public:
+  AverageArray(int n, int e) : n(n), e(e), c(0), valid(false) {
```
furushchev: @k-okada This warns order of initializer , to suppress this warning initializer should be

```diff
- AverageArray(int n, int e) : n(n), e(e), c(0), valid(false) {
+ AverageArray(int n, int e) : valid(false), n(n), e(e), c(0) {
```

This will be fixed at the next commit.

jsk_perception/src/draw_labeled_rect.cpp
furushchev: @wkentaro Sorry for late reply. I of course tried to use this function and this works but not so brilliant to distinguish colors. Current implementation is taken care of providing different colors for each neighbor class numbers.

jsk_perception/src/draw_labeled_rect.cpp
furushchev: @wkentaro this function returns color for each label. renamed to `getLabelColor`

jsk_perception/src/draw_labeled_rect.cpp
furushchev: @wkentaro Thanks. updated.

jsk_perception/src/draw_labeled_rect.cpp
```diff
@@ -0,0 +1,147 @@
...
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2015, JSK Lab
```
furushchev: @k-okada Thank you updated!

jsk_perception/src/draw_labeled_rect.cpp
furushchev: @wkentaro 

> Why not name it as `draw_rects.cpp` or `draw_rect_array.cpp` with subscription of `~input/rect_array` and `~input/labeled_rect_array`, with adjustments of code for both `RectArray` and `LabeledRectArray`?

OK. updated and now node is `draw_rects.cpp` which subscribes `RectArray` and `LabeledRectArray`

jsk_perception/src/draw_labeled_rect.cpp
wkentaro: I see.

jsk_perception/CMakeLists.txt
furushchev: @k-okada Updated image conversion from rgb8 uchar to float32 referring to your codes and it looks that it works well now. Thanks!

jsk_perception/CMakeLists.txt
```diff
@@ -258,6 +261,8 @@ add_library(${PROJECT_NAME} SHARED ${jsk_perception_nodelet_sources}
   slic/slic.cpp
   src/histogram_of_oriented_gradients.cpp
   )
+# add definitions for darknet to keep compatibility to darknet library
+add_darknet_definitions(${PROJECT_NAME})
```
furushchev: @k-okada Updated!

jsk_perception/launch/darknet_yolo_object_segmentation.launch
```diff
@@ -0,0 +1,12 @@
+<launch>
+  <arg name="input_image" default="/camera/rgb/image_rect_color" />
+
+  <node name="darknet_yolo_object_detection"
+        pkg="jsk_perception" type="darknet_object_detection"
+        output="screen">
+    <param name="network_weight_path" value="$(find darknet)/weights/yolo.weights" />
+    <param name="network_config_path" value="$(find darknet)/cfg/yolo.cfg" />
+    <param name="label_path" value="$(find darknet)/data/coco.names" />
```
furushchev: @k-okada Could you try with the latest revison and check this problem still occurs? In my environment this error does not show up.
And, what do you mean `other than add label directory under ~/.ros/data? `

furushchev:
@k-okada @wkentaro 
I pushed some additional commits, and now I think all problems discussed above was solved.
Please let me know if there is anything I overlooked or any other problems.

k-okada:
@wkentaro any comments?

jsk_perception/launch/darknet_yolo_object_segmentation.launch
```diff
@@ -0,0 +1,12 @@
+<launch>
```
wkentaro: Name should be `darkent_yolo_object_detection`.

jsk_perception/sample/sample_darknet_object_detection.launch
```diff
@@ -0,0 +1,29 @@
+<launch>
+  <arg name="launch_camera" default="true" />
```
wkentaro: Could you please `launch_camera:=false` as default?
Most of the sample launch in jsk_perception is standalone.

jsk_recognition_msgs/msg/Label.msg
```diff
@@ -0,0 +1,3 @@
+int32 index
```
wkentaro: Same comments as in https://github.com/jsk-ros-pkg/jsk_recognition/pull/1981/files#r91411074

wkentaro:
It looks really nice work.
@furushchev Could you please add test as well?
Maybe you can refer to https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/test/add_mask_image.test.

PR #1762, opened 2016-06-24
Align bounding boxes with target frame in ClusterPointIndicesDecomposer

wkentaro:
For https://github.com/start-jsk/jsk_apc/pull/1794

PR #1829, opened 2016-08-07
Organize CMakeLists.txt of jsk_perception

wkentaro:
- Refactor: jsk_perception_add_rostest -> _add_rostest
- Refactor: jsk_perception -> ${PROJECT_NAME}
- Refactor: jsk_perception_nodelet -> _add_nodelet
- Sort service files
- Fix if block syntax
- Fix missing CATKIN_DEPENDS of posedetection_msgs
- Fix node executables installation by introducing macro
- Organize cmake setup order

PR #1986, opened 2016-12-13
Fix dependency of jsk_pcl_ros_utils

wkentaro:
Fix https://github.com/jsk-ros-pkg/jsk_recognition/pull/1762#issuecomment-266618656


k-okada:
https://github.com/jsk-ros-pkg/jsk_recognition/pull/1829#issuecomment-266659307
> Maybe it is a bug, and the reason why the test pass is jsk_perception depends on jsk_data and rosdep installs it.

I see, thanks.

PR #1989, opened 2016-12-16
No longer required python-gdown dependency

wkentaro:
Because python-gdown-pip is installed via jsk_data


k-okada:
wait until new jsk_data is released

k-okada:
released https://github.com/jsk-ros-pkg/jsk_common/commit/5127e0bc8e9e88b827a19a70c70013cbc66283ba

PR #1995, opened 2017-02-03
[jsk_pcl_ros/organized_edge_detector] wait initialization

mmurooka:
sometimes (randomly) I get the following error, and that's because the subscribe callback function is called before initialization.
```
[FATAL] [1486125390.124755479]: unknown estimation method: 26639744
OpenCV Error: Assertion failed (s >= 0) in setSize, file /build/buildd/opencv-2.4.8+dfsg1/modules/core/src/matrix.cpp, line 116
terminate called after throwing an instance of 'cv::Exception'
  what():  /build/buildd/opencv-2.4.8+dfsg1/modules/core/src/matrix.cpp:116: error: (-215) s >= 0 in function setSize
```

@wkentaro , cloud you review this?

### DIFF
--- a/jsk_pcl_ros/src/organized_edge_detector_nodelet.cpp
+++ b/jsk_pcl_ros/src/organized_edge_detector_nodelet.cpp
@@ -49,6 +49,8 @@ namespace jsk_pcl_ros
 {
   void OrganizedEdgeDetector::onInit()
   {
+    boost::mutex::scoped_lock lock(mutex_);
+
     ConnectionBasedNodelet::onInit();
 
     ////////////////////////////////////////////////////////


furushchev:
@mmurooka I think you can use `boost::mutex::scoped_lock` for such a critical process. how about add class member variable `boost::mutex mutex_;` and add `boost::mutex::scoped_lock lock(mutex_);` on the first line of `onInit` and callback functions?

wkentaro:
> @mmurooka I think you can use boost::mutex::scoped_lock for such a critical process. how about add class member variable boost::mutex mutex_; and add boost::mutex::scoped_lock lock(mutex_); on the first line of onInit and callback functions?

That seems better.

mmurooka:
Thank you! updated.

wkentaro:
+1

mmurooka:
I test this PR and this did not work.
I tried a few times and always got the following error.
```
[ WARN] [1486133492.285633290, 1468919016.006140288]: [jsk_topic_tools::ConnectionBasedNodelet::warnNeverSubscribedCallback] '/edge_detector' subscribes topics only with child subscribers.
```
When I revert, it worked.

Do you have any idea?

mmurooka:
I understand the reason.
mutex is called in configCallback, which is called in onInit()...
https://github.com/mmurooka/jsk_recognition/blob/d1d70d626560b6b066318c1fb56504ad5b3ef73a/jsk_pcl_ros/src/organized_edge_detector_nodelet.cpp#L158

furushchev:
@mmurooka I'd recommend to move initialization of publishers after https://github.com/mmurooka/jsk_recognition/blob/d1d70d626560b6b066318c1fb56504ad5b3ef73a/jsk_pcl_ros/src/organized_edge_detector_nodelet.cpp#L102 and lock after dynamic reconfigure setup.

mmurooka:
make new PR https://github.com/jsk-ros-pkg/jsk_recognition/pull/1996 .
I think the cause of the first (original) error is that the topic callback is called before configCallback is called.
So I think mutex is no longer necessary.

PR #2009, opened 2017-02-06
Fix edge distance calculation in duplication check of edge depth refinement

mmurooka:
This is very complicated bug...
Before PR, the distance of edges are calculated from the distance of nearest points in lines (line has no limitation of length). In that case, two nearly-parallel but little-tilting lines are sometimes calculated to be very near even if they are not.
This PR considers the length of edges, so this bug is fixed.

- Before PR
before_pr_edited
- After PR
after_pr_edited

### DIFF
--- a/jsk_recognition_utils/include/jsk_recognition_utils/geo/segment.h
+++ b/jsk_recognition_utils/include/jsk_recognition_utils/geo/segment.h
@@ -53,6 +53,7 @@ namespace jsk_recognition_utils
     virtual double distance(const Eigen::Vector3f& point) const;
     virtual double distance(const Eigen::Vector3f& point, Eigen::Vector3f& foot_point) const;
     virtual bool intersect(Plane& plane, Eigen::Vector3f& point) const;
+    virtual void midpoint(Eigen::Vector3f& output) const;
     //virtual double distance(const Segment& other);
     friend std::ostream& operator<<(std::ostream& os, const Segment& seg);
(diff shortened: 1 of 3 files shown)


jsk_recognition_utils/include/jsk_recognition_utils/geo/segment.h
```diff
@@ -53,6 +53,7 @@ namespace jsk_recognition_utils
     virtual double distance(const Eigen::Vector3f& point) const;
     virtual double distance(const Eigen::Vector3f& point, Eigen::Vector3f& foot_point) const;
     virtual bool intersect(Plane& plane, Eigen::Vector3f& point) const;
+    virtual void midpoint(Eigen::Vector3f& output) const;
```
wkentaro: LGTM, but maybe `output` -> `midpoint` is better?

mmurooka:
Thank you for review and merge.

jsk_recognition_utils/include/jsk_recognition_utils/geo/segment.h
k-okada: @mmurooka forget to include this change, please send as another pR

mmurooka: fixed at https://github.com/jsk-ros-pkg/jsk_recognition/pull/2013

PR #2018, opened 2017-02-11
[jsk_recognition_utils] add default constructor to geometry classes.

mmurooka:
If I understand C++ correctly, this PR is necessary to use instance of jsk_recognition_utils/geometry classes as member variable of the other class.

For example, the following code becomes error without this PR.
https://github.com/mmurooka/euslib/blob/murooka-branch/demo/murooka/wholebody_manipulation_planner/em_motion_estimation/include/em_motion_estimation/pcl/point_element.h#L16-L17


YoheiKakiuchi:
@mmurooka 
Is this the same as https://github.com/jsk-ros-pkg/jsk_recognition/pull/2030 ?
Can you close this PR?

mmurooka:
I forgot here, too...

PR #2019, opened 2017-02-12
[jsk_recognition_utils/geo/Polygon] add distance method.

mmurooka:
compute distance to point.

Line segment version is here.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_recognition_utils/src/geo/segment.cpp#L77-L88


YoheiKakiuchi:
This is the same as https://github.com/jsk-ros-pkg/jsk_recognition/pull/2031

mmurooka:
Oh, I forgot this PR and make same PR!

PR #2025, opened 2017-03-08
[jsk_perception/apply_mask_image] Add negative option

iory:
original image
image_publisher

mask image
mask_publisher

apply_mask image
apply_mask_image

apply_negative_mask 
apply_negative_mask_image

### DIFF
--- a/jsk_perception/src/apply_mask_image.cpp
+++ b/jsk_perception/src/apply_mask_image.cpp
@@ -48,6 +48,8 @@ namespace jsk_perception
     DiagnosticNodelet::onInit();
     pnh_->param("approximate_sync", approximate_sync_, false);
     pnh_->param("clip", clip_, true);
+    pnh_->param("negative", negative_, false);
+    pnh_->param("negative/before_clip", negative_before_clip_, true);
     pnh_->param("mask_black_to_transparent", mask_black_to_transparent_, false);
     pnh_->param("queue_size", queue_size_, 100);
     pub_image_ = advertise<sensor_msgs::Image>(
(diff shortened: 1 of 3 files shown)


wkentaro:
how to handle clip with negative?

iory:
in https://github.com/iory/jsk_recognition/blob/f579faff74d5b68fb529fdf8823dd2b37b184b34/jsk_perception/src/apply_mask_image.cpp#L110-L119,
```
    if (negative_) {
      cv::bitwise_not(mask, mask);
    }

    if (clip_) {
      cv::Rect region = jsk_recognition_utils::boundingRectOfMaskImage(mask);
      mask = mask(region);
      image = image(region);
    }
```
Since processing to make it negative is before clip, there is no problem.
This option is intended to absorb the differences between the cases as follows.
For example, a person who wishes to retrieve a part of the robot can accomplish by applying apply_mask_image to robot_self_filter, but for a person wanting to extract the part of the robot, that negative is necessary.
mask_publisher
image_publisher
apply_mask_image
apply_negative_mask_image

wkentaro:
I see. But usually if `negative=True` and `clip=True` ends up no clipping because clip finds the bounding box of white region.
I'm not sure this is okay.

iory:
OK, add nagative_before_clip option. 
this solves order of clipping and negative

jsk_perception/src/apply_mask_image.cpp
```diff
@@ -105,11 +107,25 @@ namespace jsk_perception
...
+    } else {
+      if (clip_) {
+        cv::Rect region = jsk_recognition_utils::boundingRectOfMaskImage(mask);
+        mask = mask(region);
+        image = image(region);
+      }
+      if (negative_) {
+        cv::bitwise_not(mask, mask);
+      }
     }
```
wkentaro: Why not below?

```bash
if (negative_before_clip_ && negative_) {
  cv::bitwise_not(...)
}

if (clip_) {
  # clip
}

if (!negative_before_clip_ && negative_) {
  cv::bitwise_not(...)
}
```

iory: Changed

wkentaro:
LGTM

wkentaro:
Ah, wait. Maybe you should add some descriptions about the new parameters in doc @iory.

iory:
OK, I add

wkentaro:
LGTM

PR #2029, opened 2017-03-13
Add PointCloudXYZRGBToXYZ

wkentaro:
## Application

PointCloudXYZRGB -> [PointCloudXYZRGBToXYZ] -> PointCloudXYZ = A
Label -> [ColorizeLabel] = B
(A, B) -> [AddColorFromImage] -> LabeledPointCloud = C

**B**

frame0000

**C**

### DIFF
--- a/doc/jsk_pcl_ros_utils/nodes/pointcloud_xyz_to_xyzrgb.md
+++ b/doc/jsk_pcl_ros_utils/nodes/pointcloud_xyz_to_xyzrgb.md
@@ -15,3 +15,9 @@ Node to convert fields of `sensor_msgs/PointCloud2` from `XYZ` to `XYZRGB`.
 * `~output` (`sensor_msgs/PointCloud2`)
 
   Output cloud whose field is `XYZRGB`.
+
+## Sample
+
+```bash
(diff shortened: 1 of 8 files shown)


knorth55:
+1
But i think you can use PointCloudXYZRGBtoXYZI in pcl.
Why do you prefer PointCloudXYZ to PointCloudXYZI?

wkentaro:
What is I?

wkentaro:
I've found AddColorFromImage converts PointXYZRGB to PointXYZ at the initial point.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/src/add_color_from_image_nodelet.cpp#L87
So currently this node is not necessary for me.

wkentaro:
@knorth55 Thanks for your review.

knorth55:
I is intensity and it is for gray scale image.
I don't know which format  XYZ or XYZI is better, but pcl default conversion is XYZRGB to XYZI, so i thought youcan use the convert function for this nodelet.

wkentaro:
I see. Thank you

PR #2043, opened 2017-03-27
[jsk_pcl_ros_utils] ensure super class functionality works

furushchev:
Some nodelet forgot to enable super class functionality:

- Forgot to call `vital_checker_->poke()`

Without calling this method on callback, runtime_monitor displays error state on the nodelet and it is very confusing on debugging.

- Forgot to call `onInitPostProcess()`

Without calling this method on `onInit()` method, nodelet does not subscribe on start even if param `always_subscribe` is set to `true`.

### DIFF
--- a/jsk_pcl_ros_utils/src/add_point_indices_nodelet.cpp
+++ b/jsk_pcl_ros_utils/src/add_point_indices_nodelet.cpp
@@ -44,6 +44,7 @@ namespace jsk_pcl_ros_utils
     DiagnosticNodelet::onInit();
     pnh_->param("approximate_sync", approximate_sync_, false);
     pub_ = advertise<PCLIndicesMsg>(*pnh_, "output", 1);
+    onInitPostProcess();
   }
 
   void AddPointIndices::subscribe()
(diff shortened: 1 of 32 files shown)


k-okada:
can we call this method at
https://github.com/jsk-ros-pkg/jsk_common/blob/2.2.3/jsk_topic_tools/src/diagnostic_nodelet.cpp#L63

--
◉ Kei Okada


2017-03-27 23:44 GMT+09:00 Furushchev <notifications@github.com>:

furushchev:
@k-okada Unfortunately, I think no, we should wait calling `subscribe` until all initialization is done.

wkentaro:
> Unfortunately, I think no, we should wait calling subscribe until all initialization is done and we have no way to hook on onInit is done.

Yeah, we need to wait the advertising in `onInit` method.

k-okada:
ok, merged, but this could be very dangerous, like `init-ending` of euslisp robot model. i'm afraid a lot of people will forget to add this when they create new nodelet.

PR #2062, opened 2017-04-19
FilterMaskImageWithSize: Filter mask image with its size

wkentaro:
Modified:
- jsk_perception/CMakeLists.txt
- jsk_perception/include/jsk_perception/multiply_mask_image.h
- jsk_perception/plugins/nodelet/libjsk_perception.xml Added:
- jsk_perception/cfg/FilterMaskImageWithSize.cfg
- jsk_perception/include/jsk_perception/filter_mask_image_with_size.h
- jsk_perception/src/filter_mask_image_with_size.cpp

filter_mask_image_with_size


k-okada:
@wkentaro I have resolved conflict via web interface, so please confirm if that is ok.

wkentaro:
Thanks. I rebased.

knorth55:
I approved, but i don't see usecase of this nodelet. (Sample launch subscribe same mask image.)

wkentaro:
I added flag `~use_reference` to minimize the overhead caused by synchronizing.

wkentaro:
@knorth55 Could u Please check the latest commit?

wkentaro:
Please merge this.

k-okada:
You can also use back projection method for detection motion area , if the camera is fixed on world frame-> https://github.com/ros-perception/opencv_apps/pull/62
note that a camera with automatic white balance features does not work well with that algorithm, you'd better to use good camera.

And if we can assume the moving object is human, we can use that for human detection application -> http://goldberg.berkeley.edu/pubs/acc-2012-visual-tracking-final.pdf
```
This algorithm combines statistical background image estimation and per-pixel Bayesian segmentation. It was introduced by Andrew B. Godbehere, Akihiro Matsukawa, Ken Goldberg in their paper “Visual Tracking of Human Visitors under Variable-Lighting Conditions for a Responsive Audio Art Installation” in 2012. As per the paper, the system ran a successful interactive audio art installation called “Are We There Yet?” from March 31 - July 31 2011 at the Contemporary Jewish Museum in San Francisco, California.
```

wkentaro:
> You can also use back projection method for detection motion area , if the camera is fixed on world frame-> ros-perception/opencv_apps#62
note that a camera with automatic white balance features does not work well with that algorithm, you'd better to use good camera. > @wkentaro

Thank you for the info.

> And if we can assume the moving object is human, we can use that for human detection application

I tried using [fcn_object_segmentation.py](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/fcn_object_segmentation.html) for person detection for object data collection, but it was a bit difficult to segment the person and him/her holding objects.

PR #2090, opened 2017-05-15
Cleanup octomap dependencies of jsk_pcl_ros

wkentaro:
Following fix in https://github.com/jsk-ros-pkg/jsk_recognition/issues/2087

### DIFF
--- a/jsk_pcl_ros/package.xml
+++ b/jsk_pcl_ros/package.xml
@@ -54,7 +54,6 @@
   <build_depend>octomap</build_depend>
   <build_depend>octomap_server</build_depend>
   <build_depend>octomap_msgs</build_depend>
-  <build_depend>octomap_ros</build_depend>
   <build_depend>jsk_pcl_ros_utils</build_depend>
   <build_depend>jsk_data</build_depend>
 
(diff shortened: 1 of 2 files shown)


jsk_pcl_ros/package.xml
```diff
@@ -110,8 +109,6 @@
   <run_depend>nav_msgs</run_depend>
   <run_depend>octomap</run_depend>
   <run_depend>octomap_server</run_depend>
-  <run_depend>octomap_msgs</run_depend>
-  <run_depend>octomap_ros</run_depend>
```
wkentaro: octomap_ros is not needed and there seems no dependency.

jsk_pcl_ros/package.xml
```diff
@@ -110,8 +109,6 @@
   <run_depend>nav_msgs</run_depend>
   <run_depend>octomap</run_depend>
   <run_depend>octomap_server</run_depend>
-  <run_depend>octomap_msgs</run_depend>
```
wkentaro: `octomap_msgs` has no dynamic link, so it is just a build_depend.

wkentaro:
Sorry for making you confused.
No, this is just a refactoring.

We still need octomap_ros from source if octomap is from-source,
because `jsk_pcl_ros` explicitly depends on `octomap_server`, `octomap_server` depends on `octomap_ros`, in addition, `octomap_ros`, `octomap_server` and `jsk_pcl_ros` depends on `octomap`.

mmurooka:
I now understand. Thank you.

wkentaro:
Could this be merged?

PR #2094, opened 2017-05-26
[jsk_pcl_ros][ColorHistogramMatcher] add skip_compare option

furushchev:
I'd like to only use the result of processing (histogram) and not to match with any references.
If `skip_compare` is set to `true`, matching will be disabled.


wkentaro:
I think it is better that we create another node which computes color histogram.
What do you think about this? @furushchev

furushchev:
@wkentaro Agree. Comming soon! :+1:

PR #2102, opened 2017-06-02
Create rgb image in depth_image_creator

wkentaro:
## Why?

I need to get rgb image seen from different view point for #2104 .

sample_depth_image_creator

cc @YutoUchimi


k-okada:
it seems some pixel of created rgb/depth image is black, is it ok? do we loop over image coordinates, not point cloud list?
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2102/files#diff-ace21927127e2a29fdb36ab72b35676cR259

wkentaro:
I think it is ok, the black region has value of nan for depth and 0 for rgb image.
No, we loop over point cloud list.

PR #2112, opened 2017-06-12
[doc/jsk_pcl_ros] modify PointcloudScreenpoint

kochigami:
- modified one of publishing topic names based on 
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/src/pointcloud_screenpoint_nodelet.cpp#L73

- modified one of subscribing topic names based on
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/src/pointcloud_screenpoint_nodelet.cpp#L86 

I also added missing ```:``` .

If there is any problem, please let me know.


k-okada:
Thank you for your contributions

PR #2125, opened 2017-06-21
[jsk_perception/polygon_to_mask] add error message of frame_id

mmurooka:
I took 20 minutes to find this mistake, so add error message.


mmurooka:
@wkentaro could you review ?

PR #2129, opened 2017-06-22
Rewrite KinfuNodelet with some enhancements and new features

wkentaro:
- [x] Stable tracking (thanks to @mmurooka but we don't know why)
- [x] Cloud output (thanks to @mmurooka )
- [x] Rendered image (thanks to @mmurooka ) (*feature*)
- [x] Mesh
- [x] Color integration (*feature*)
- [x] Colorized rendered image (*feature*)
- [x] Generated depth image (*feature*)
- [x] Colorized cloud output (*feature*)
- [x] Texture mesh (*feature*)
- [x] SLAM

## Sample

```bash
roslaunch jsk_pcl_ros sample_kinfu.launch
rosservice call /kinfu/save_mesh  # saved as ~/.ros/jsk_pcl_ros/sample_kinfu/mesh.obj
rosservice call /kinfu/reset
```

sample_kinfu_x0 4

**Mesh**

cc @mmurooka


wkentaro:
**Colorized rendered image**

sample_kinfu_002_x0 4

wkentaro:
**generated depth**


sample_kinfu_003_x0 4

wkentaro:
**Colorized cloud output**

sample_kinfu_004_x0 4

wkentaro:
**Texture mesh**

snapshot00

wkentaro:
@YutoUchimi Now we can generate texture mesh model by KinfuNodelet. Please try this branch with `jsk_pcl/AttentionClipper` to generate object model used in ARC2017.

mmurooka:
All source code is rewritten in this PR.
I'd like to add cc to all of this nodelet users. Do you know other users?

Please list up what feature was removed (for example odom tf).
Also, are topic name and service name changed?
launch/kinfu.launch no longer work, so you need to remove.

wkentaro:
## Topic / service change

- `~input/color` (same)
- `~input/depth` (same)
- `~output` (removed)
- `~output/cloud` -> `~output`
- `~save_mesh` (samve)
- `~reset` (new)

## Behavior change

- auto reset -> reset by service and add flag of `~auto_reset`
- subscribe color -> won't require but add flag of `~integrate_color`
- save mesh.ply -> save mesh.obj

wkentaro:
## TODO

- [ ] Fix kinfu.launch
- [ ] Support kinfu for slam

wkentaro:
sample_kinfu_006
https://drive.google.com/file/d/0B9P1L--7Wd2vTEZod3EtSDUxcEk/view?usp=sharing

mmurooka:
Good. I'll see this PR tomorrow. Please tell when you finish updating this PR.

wkentaro:
It seems https://github.com/jsk-ros-pkg/jsk_recognition/pull/2129/commits/1923e23de596750a3509d4e43071e38fe3fa7bc6 does not work well, because of the unneeded tf resolve from odom (fixed_frame) to camera.
So I think static_tf_publisher or some dynamic tf publisher should be used to improve the view on rviz.

wkentaro:
(Update)
**Kinfu for slam** (Publish map -> odom_init tf)

image

https://drive.google.com/file/d/0B9P1L--7Wd2vVlMwNjJNeGVIV0E/view

wkentaro:
Finished. I added documentation also:

doc001
doc002
doc003

wkentaro:
A possible improvement is parameterize "odom_init" with `~fixed_frame_id`. Do you have any idea? @mmurooka

jsk_pcl_ros/src/kinfu_nodelet.cpp
```diff
@@ -33,236 +32,547 @@
...
-        pub_cloud_.publish(ros_cloud);
       }
-      catch (...) {
-        NODELET_ERROR("Failed to lookup transform from %s to %s",
-                          child_frame_id_.c_str(),
-                          depth_image->header.frame_id.c_str());
-        return;
+    }
+
+    // publish generated depth image
```
mmurooka: I think that generated depth in necessary only in limited use case. So how about adding rosparam to enalbe it and setting disabled as default?

jsk_pcl_ros/src/kinfu_nodelet.cpp
```diff
@@ -33,236 +32,547 @@
...
+        pcl::toROSMsg(*cloud_xyz, output_cloud_msg);
+      }
+      output_cloud_msg.header.frame_id = "kinfu_origin";
+      pub_cloud_.publish(output_cloud_msg);
     }
   }
-  bool Kinfu::saveMeshService(std_srvs::Empty::Request& req, std_srvs::Empty::Response& res)
+
+  bool
+  Kinfu::resetCallback(std_srvs::Empty::Request& req, std_srvs::Empty::Response& res)
```
wkentaro: Yeah, you're right.

jsk_pcl_ros/src/kinfu_nodelet.cpp
wkentaro: Maybe I can check the number of subscribers.

jsk_pcl_ros/src/kinfu_nodelet.cpp
```diff
@@ -33,236 +32,547 @@
...
+    float vsz = pcl::device::kinfuLS::VOLUME_SIZE;
+    float shift_distance = pcl::device::kinfuLS::DISTANCE_THRESHOLD;
+    Eigen::Vector3f volume_size = Eigen::Vector3f::Constant(vsz/*meters*/);
+    if (shift_distance > 2.5 * vsz)
+    {
+      NODELET_WARN("WARNING Shifting distance (%.2f) is very large compared to the volume size (%.2f).",
+                   shift_distance, vsz);
+    }
+
+    kinfu_ = new pcl::gpu::kinfuLS::KinfuTracker(volume_size, shift_distance, caminfo_msg->height, caminfo_msg->width);
```
wkentaro: Why do you think so?

wkentaro:
> Old nodelet can change kinfu parameters such as tsdf_trunc_dist_. I've never tried kinfu with changing such parameter, so I don't know such feature is effective. How do you think?

I think it is better to set these parameters `onInit`, because it requires many `mutex` handling when we use `kinfu_` in class methods.
Currently I don't encounter the situation when I want to change the parameter. If needed, I will do that.

> What is the application of colored texture generation and rendered depth image?

- colored texture generation -> to generate textured mesh model of objects
- rendered depth image -> get organized depth/pointcloud to integrate with image-based recognition results (rendered depth looks better than the sensor input).

jsk_pcl_ros/src/kinfu_nodelet.cpp
mmurooka: I don't know C++ pointer well, but I think if kinfu is reset N times, new KinfuTracker object is generated N times without releasing memory. Is it my incorrect understanding?

jsk_pcl_ros/src/kinfu_nodelet.cpp
mmurooka: Yes, it is OK.

wkentaro:
## TODO

- [ ] Visualize `odom_init`, `map` tf in slam movie
- [ ] Hz of colored/non-colored output cloud
- [ ] Hz of generated_depth

mmurooka:
> Currently I don't encounter the situation when I want to change the parameter. If needed, I will do that.

I think changing kinfu parameter during running kinfu may be minor usage, so I think it is not necessary.

jsk_pcl_ros/src/kinfu_nodelet.cpp
```diff
@@ -33,236 +32,547 @@
...
+        // use kinfu as slam, and publishes tf: map -> fixed_frame_id_ (usually odom_init)
+        try
+        {
+          tf::StampedTransform tf_odom_to_camera;
+          tf_listener_->lookupTransform(
+            fixed_frame_id_, caminfo_msg->header.frame_id, ros::Time(0), tf_odom_to_camera);
+          Eigen::Affine3f odom_to_camera;
+          tf::transformTFToEigen(tf_odom_to_camera, odom_to_camera);
+
+          Eigen::AngleAxisf rotate90_x = Eigen::AngleAxisf(1.57, Eigen::Vector3f::UnitX());
```
mmurooka: I double checked the slam transformation formula and this implementation should be correct without rotate90_x.
Did you need this with rosbag test?

wkentaro: I needed this to make rviz scene good to see.
Without this, the map is upside-down.

wkentaro:
> Publishing Kinfu status (succeeded or lost) will be useful.
Topic message contains header and bool.
mmurooka/euslib@aabe84c
How do you think?

Yeah, that would be great.  I will use [`jsk_recognition_msgs/TrackerStatus`](https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_recognition_msgs/msg/TrackerStatus.msg).

mmurooka:
> Hz of generated_depth

It will be OK even if generated_depth is slow because high-quality depth will be useful in some cases even it is slow. I worry that kinfu cloud topic becomes slow because of generated_depth even if it's not necessary. https://github.com/jsk-ros-pkg/jsk_recognition/pull/2129#discussion_r123880519 will solve this.

jsk_pcl_ros/src/kinfu_nodelet.cpp
mmurooka: It's strange that map is upside-down because that means odom_init recorded in rosbag is also upside down. Watching odom_init frame in rosbag will give hint.

jsk_pcl_ros/src/kinfu_nodelet.cpp
wkentaro: Handled by 7ae9d12

jsk_pcl_ros/src/kinfu_nodelet.cpp
wkentaro: You're right, and handled by f79d89d

wkentaro:
Now it looks good about slam.
image
https://drive.google.com/drive/u/1/folders/0B9P1L--7Wd2vRlozbzZiZU8wYkE

mmurooka:
Good. Do you know why you need 90 deg rotation before and not need now?

wkentaro:
No.

wkentaro:
Actually `jsk_recognition_msgs/TrackerStatus` is not designed to describe **Tracking Status** but the status of tracker about which is trying to track or not.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/scripts/tracker_status_info.py#L21-L23

mmurooka:
> Now it looks good about slam.
https://drive.google.com/drive/u/1/folders/0B9P1L--7Wd2vRlozbzZiZU8wYkE

Link is directory.

wkentaro:
Updated.

mmurooka:
Please tell me when you finish all update.

wkentaro:
Ok, I removed the last commit that can be added by another PR.

wkentaro:
@mmurooka Thank you for your review! @k-okada Coud you please merge this?

PR #2131, opened 2017-06-26
Publish largest mask converted from label

knorth55:
## Why?

To select largest object as the target from label input.


wkentaro:
The largest object is recognized with https://github.com/jsk-ros-pkg/jsk_recognition/pull/2133 in a 3D way, so we don't need this and can close this PR.
The 2D way has less compatibility with AttentionClipper, and that's why we need to use 3D info.

PR #2141, opened 2017-07-03
jsk_pcl_ros: Add Primitive shape classifier nodelet

furushchev:
Nodelet to classify shapes of each cluster point clouds on planes.

- [x] add nodelet
- [x] add sample launch file
- [x] add docs
- [x] add pics
- [x] add test codes


furushchev:
Needs another review?

PR #2149, opened 2017-07-05
Support PCA even without input planes but with only ground frame

wkentaro:
The purple bounding box does not have input planes, and this is useful for bounding box-based grasp planning.

**without pca**

sample_cluster_point_indices_without_pca

**with pca**

sample_cluster_point_indices_with_pca


wkentaro:
@iory could you please review this?

708yamaguchi:
I tested this code at ARC 2017 stow task, and it worked!

with PCA
screenshot 2017-07-06 17 38 55

PR #2158, opened 2017-07-09
Check encoding of input topics in FuseImages

wkentaro:
## Why?

Currently it can be passed 16UC1 depth and 32FC1 one to be fused, but it is not possible because they have different metrics: mm for 16UC1 and m for 32FC1.

Found at https://github.com/start-jsk/jsk_apc/pull/2244


wkentaro:
@k-okada please merge this

PR #2159, opened 2017-07-10
[jsk_pcl_ros] add sample and test of sorting decomposer

knorth55:
add sample and test for `sort_by` function in `cluster_point_indices_decomposer`


wkentaro:
Because we already have this: https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/sample/sample_cluster_point_indices_decomposer_sort_by.launch

PR #2176, opened 2017-07-23
Convert bounding box to mask

wkentaro:
with @YutoUchimi 

image


knorth55:
we use this in ARC2017

wkentaro:
> we use this in ARC2017

Actually, we used https://github.com/jsk-ros-pkg/jsk_recognition/pull/2183 instead of this, because point indices does represent more precisely the segmentation than box.

PR #2177, opened 2017-07-23
Fixes on probabilistic image classifier

wkentaro:


knorth55:
we use this in ARC2017

Issue #1308, opened 2015-11-09
Segfault SLICSuperPixels because of large image


wkentaro:
https://gist.github.com/wkentaro/84e4cd9ee84a206cef3e

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffef179700 (LWP 12514)]
[New Thread 0x7fffee978700 (LWP 12515)]
[New Thread 0x7fffee177700 (LWP 12516)]
[New Thread 0x7fffed976700 (LWP 12517)]
[New Thread 0x7fffed175700 (LWP 12518)]
[New Thread 0x7fffec974700 (LWP 12519)]
[New Thread 0x7fffe7fff700 (LWP 12521)]
[New Thread 0x7fffe77fe700 (LWP 12522)]
[New Thread 0x7fffe6ffd700 (LWP 12523)]
[New Thread 0x7fffe67fc700 (LWP 12524)]
[New Thread 0x7fffe5ffb700 (LWP 12525)]
[New Thread 0x7fffe57fa700 (LWP 12526)]
[New Thread 0x7fffe4ff9700 (LWP 12527)]
[ INFO] [1447083483.356138142]: Initializing nodelet with 8 worker threads.
[New Thread 0x7fffdbfff700 (LWP 12528)]
[New Thread 0x7fffad0be700 (LWP 12533)]
[New Thread 0x7fffab4d0700 (LWP 12562)]
[New Thread 0x7fffab0cf700 (LWP 12563)]
[New Thread 0x7fffaacce700 (LWP 12565)]
[New Thread 0x7fffaa8cd700 (LWP 12566)]
[New Thread 0x7fffaa0cb700 (LWP 12569)]
[New Thread 0x7fffaa4cc700 (LWP 12567)]
[New Thread 0x7fffa9cca700 (LWP 12570)]
*** Error in `/home/wkentaro/ros/indigo/devel/lib/jsk_perception/slic_super_pixels': munmap_chunk(): invalid pointer: 0x00007fff9f5674a0 ***

Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffec974700 (LWP 12519)]
0x00007ffff6d83cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
[...]

PR #2181, opened 2017-07-27
[jsk_perception, slic_super_pixels] add parameter, publish_debug_images

YoheiKakiuchi:


wkentaro:
why not check the num of subscribers?

YoheiKakiuchi:
The ports are used for just debug purpose.
It's enough to select by parameter.

wkentaro:
So you have more than 9999 topics in your system?

YoheiKakiuchi:
> So you have more than 9999 topics in your system?
???

wkentaro:
> The ports are used for just debug purpose.

You have some trouble caused by missing ports, right?

YoheiKakiuchi:
No.
Some bugs may exist for generating debug images. 
Node crashes at random when using debug images.

wkentaro:
So, you can use try-catch?

YoheiKakiuchi:
try-catch do not work for segfault

wkentaro:
I see. Then you should add comment to the code about problems you encountered, and why you need this PR. Please refer to this PR in the comment.
Basically, switching debug publication by rosparam is not necessary, in my opinion.

wkentaro:
BTW, your segfault problem is similar to https://github.com/jsk-ros-pkg/jsk_recognition/issues/1308 which is caused by large input image size?

Issue #2186, opened 2017-08-11
[multi plane extraction] plane occasionally flips when use_sensor_frame is false


furushchev:
If use_sensor_frame is false, view point vector that is used to normal axis of plane is set without initialization. 
Without seeing viewpoint vector pcl computes it using PCA but it is not stable for flipping. 

ref:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/src/multi_plane_extraction_nodelet.cpp#L258
https://github.com/PointCloudLibrary/pcl/blob/pcl-1.7.2/segmentation/include/pcl/segmentation/impl/extract_polygonal_prism_data.hpp#L54-L67

furushchev:
Solved via #2191

PR #2191, opened 2017-08-12
jsk_pcl_ros: multi_plane_extraction: add option use_coefficients

furushchev:
Addresses #2186 

If `~use_sensor_frame` is false, the z-axis of plane may be flipped on each detection and coefficients topic is never used.
I added a new option `~use_coefficients` that determines z-axis of each planes.

**NOTE** Visualization of normal arrow in rviz is not valid. (To see actual z normal axis, we must consider both polygon and coefficients)

extraction_falure

extraction_success

### DIFF
--- a/doc/jsk_pcl_ros/nodes/multi_plane_extraction.md
+++ b/doc/jsk_pcl_ros/nodes/multi_plane_extraction.md
@@ -11,7 +11,9 @@ Extract the points above the planes between `~min_height` and `~max_height`.
 * `~input_polygons` (`jsk_recognition_msgs/PolygonArray`)
 * `~input_coefficients` (`jsk_recognition_msgs/ModelCoefficientsArray`):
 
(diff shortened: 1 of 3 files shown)


furushchev:
Just to clarify:

|                         | use_sensor_frame: false            | use_sensor_frame: true                 |   |   |
|-------------------------|------------------------------------|----------------------------------------|---|---|
| use_coefficients: false | PCL computes z-axis internally     | Determine z-axis direction to TF frame |   |   |
| use_coefficients: true  | Determine z-axis from coefficients | N/A (Warning shows up)                 |   |   |
|                         |                                    |                                        |   |   |

k-okada:
Did you update doc?

furushchev:
Totally forgot at all. Thanks.

furushchev:
@k-okada OK. Updated.

furushchev:
Now passed tests.

PR #2192, opened 2017-08-13
Support filtering bounding boxes without indices

wkentaro:


wkentaro:
@otsubo Could you please merge this?

PR #2195, opened 2017-08-14
Derive from pre-built image of ubuntu 14.04 on Jenkins

wkentaro:


wkentaro:
It seems use of docker image on docker hub is better now.

PR #2206, opened 2017-10-04
jsk_perception: people_pose_estimation_2d.py: unsynchronize camera info

### LINKED ISSUE #2165: People Pose Estimation 2D uses large memory
I'm running `people_pose_estimation_2d.py` with CUDA 6.5.
It works well on start (publish processed results), but after a few seconds it does not publish any message.

figure_1

I found that after callback is called sometimes, memory (not GPU memory) is run out.

- Why this program uses large memory (up to 2GB)
- Is it possible to reduce memory usage?

What I did are:

- ensure `del` for all chainer variables which are allocated by `.to_cpu` method
- call `gc.collect()` on every callback function 

But still no luck.
Thank you very much in advance.

furushchev:
This pull request disables synchronization of camera info with images and let the node use it only once instead.
Fixes #2165

### DIFF
--- a/jsk_perception/node_scripts/people_pose_estimation_2d.py
+++ b/jsk_perception/node_scripts/people_pose_estimation_2d.py
@@ -100,6 +100,7 @@ class PeoplePoseEstimation2D(ConnectionBasedTransport):
         self._load_model()
         self.image_pub = self.advertise('~output', Image, queue_size=1)
         self.pose_pub = self.advertise('~pose', PeoplePoseArray, queue_size=1)
+        self.sub_info = None
         if self.with_depth is True:
             self.pose_2d_pub = self.advertise('~pose_2d', PeoplePoseArray, queue_size=1)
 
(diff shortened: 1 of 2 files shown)


furushchev:
Build failed on kinetic: it took 59min on jenkins.

furushchev:
Test passed.
@iory Any comment?

jsk_perception/node_scripts/people_pose_estimation_2d.py
```diff
@@ -127,9 +127,9 @@ def subscribe(self):
                 '~input', Image, queue_size=1, buff_size=2**24)
             sub_depth = message_filters.Subscriber(
                 '~input/depth', Image, queue_size=1, buff_size=2**24)
-            sub_info = message_filters.Subscriber(
-                '~input/info', CameraInfo, queue_size=1, buff_size=2**24)
-            self.subs = [sub_img, sub_depth, sub_info]
+            self.sub_info = rospy.Subscriber(
+                '~input/info', CameraInfo, self._cb_cam_info)
+            self.subs = [sub_img, sub_depth]
```
k-okada: since this is workaround and if upstream problem was solved, we can use `message_filters` so, please un-comment old implementation, instead of removing it and add more information and link to upstream issue report.

c.f. https://github.com/jsk-ros-pkg/jsk_recognition/issues/2165#issuecomment-334919526

furushchev: @k-okada True. Added comment and option `~sync_camera_info` so we can switch on/off this feature.

PR #2213, opened 2017-11-01
Trigger the build of jsk-docs.rtd.org by travis

wkentaro:
This enables us to remove cron job from jenkins.


wkentaro:
I'm not sure why, but currently there is no run of `after_success` in travis job.
screen shot 2017-11-01 at 20 22 35

k-okada:
I suppose it will not run in forked repo.

PR #2219, opened 2017-11-18
Use Travis as much as possible

wkentaro:


wkentaro:
I found this is not possible because of long build time of jsk_recognition.

PR #2234, opened 2017-12-18
Fix uninitialized pointer error in some recognition nodelets

orikuma:
Problem:
When a `jsk_topic_tools/standalone_complexed_nodelet` loaded `jsk_pcl/tilt_laser_listener` and `resized_image_transport/ImageResizer` in the situation where input point cloud and image topics are already published, it died with following error message.
```cpp
standalone_complexed_nodelet: /usr/include/boost/smart_ptr/shared_ptr.hpp:653: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = jsk_topic_tools::VitalChecker; typename boost::detail::sp_member_access<T>::type = jsk_topic_tools::VitalChecker*]: Assertion `px != 0' failed.
```

Proposed solution:
This error seems to be caused by initialization order of the vital checker (the pointer for a vital checker is referred in a callback function before it is initialized).
This PR prevents the pointers of the vital checker from being referred before initialization has finished.


k-okada:
Thank you

PR #2244, opened 2018-02-28
Add fcn_depth_prediction node

YutoUchimi:
Add node for depth image prediction for transparent objects (i.e. objects whose depth is almost insensible by RGB-D camera).


jsk_perception/node_scripts/fcn_depth_prediction.py
```diff
@@ -0,0 +1,520 @@
...
+        else:
+            sync = message_filters.TimeSynchronizer(
+                fs=self.subs, queue_size=queue_size)
+        sync.registerCallback(self._cb)
+
+    def unsubscribe(self):
+        for sub in self.subs:
+            sub.unregister()
+
+    def colorize_depth(self, depth, min_value=None, max_value=None):
```
wkentaro: maybe it can be a standalone function instead of a class method.

jsk_perception/node_scripts/fcn_depth_prediction.py
```diff
@@ -0,0 +1,520 @@
...
+        proba_msg.header = img_msg.header
+        self.pub_proba.publish(proba_msg)
+        depth_msg = br.cv2_to_imgmsg(depth_pred.astype(np.float32))
+        depth_msg.header = img_msg.header
+        self.pub_depth.publish(depth_msg)
+        depth_raw_msg = br.cv2_to_imgmsg(depth_pred_raw.astype(np.float32))
+        depth_raw_msg.header = img_msg.header
+        self.pub_depth_raw.publish(depth_raw_msg)
+
+    def depth_predict(self, bgr, depth_bgr=None):
```
wkentaro: maybe `predict_depth` is better.

jsk_perception/node_scripts/fcn_depth_prediction.py
```diff
@@ -0,0 +1,520 @@
...
+        if self.backend == 'chainer':
+            return self._depth_predict_chainer_backend(bgr, depth_bgr)
+        raise ValueError('Unsupported backend: {0}'.format(self.backend))
+
+    def _depth_predict_chainer_backend(self, bgr, depth_bgr=None):
```
wkentaro: maybe `predict_depth` is better.

knorth55:
this test took long time, so maybe it is better to disable the test.

jsk_perception/node_scripts/fcn_depth_prediction.py
YutoUchimi: Thanks for review!
Updated.

jsk_perception/node_scripts/fcn_depth_prediction.py
YutoUchimi: Updated.

jsk_perception/node_scripts/fcn_depth_prediction.py
YutoUchimi: Updated.

YutoUchimi:
Hmm... Actually, test often fails because of timeout.

@knorth55 
Why do you suggest disabling the test?
To avoid failing because of this PR in another PR?

knorth55:
>Hmm... Actually, test often fails because of timeout.

This is the reason why I suggest to disable the test.
This test actually requires GPU, but jenkins run the test on CPU, which causes timeout.

YutoUchimi:
OK. Removed test commit.

PR #2256, opened 2018-04-03
Improve visualization in tile_image.py

wkentaro:
- Use FONT_HERSHEY_SIMPLEX.
- Adjust font_scale according to the new font.

**Before**

**After**


wkentaro:
@k-okada could you merge this?

PR #2257, opened 2018-04-03
Add MaskRCNNInstanceSegmentation node

wkentaro:
**NOTE**: This model successfully detects the instance masks of two persons in this example.

mask_rcnn_instance_segmentation

It runs in ~10fps with Titan X Pascal.

cc: @chibi314 @YoheiKakiuchi


jsk_perception/scripts/install_trained_data.py
```diff
@@ -163,6 +163,14 @@ def main():
         md5='a585e4d41ed67d5052417ade6fb2d608',
     )
 
+    # node_scripts/mask_rcnn_instance_segmentation.py
+    download_data(
+        pkg_name=PKG,
+        path='trained_data/mask_rcnn_resnet50_voc_20180403',
```
YutoUchimi: Probably `path='trained_data/mask_rcnn_resnet50_voc_20180403.npz',`?

wkentaro: Thanks! fixed!

knorth55:
How about publishing both `mask_indices` and `mask`, too.
I mean mask is gray scale image with `0` for `False` and `255` for `True`.

wkentaro:
> How about publishing both mask_indices and mask, too.
I mean mask is gray scale image with 0 for False and 255 for True.

You mean it should publish `(N_instances, H, W)` uint8 masks?

knorth55:
> You mean it should publish (N_instances, H, W) uint8 masks?

Yes. Is it too heavy?

wkentaro:
Yeah.
But anyway what is the advantage of that.

knorth55:
In the case we only nee mask images, we currently need to subscribe image for the image height and width.
If the node publish masks in image shape, it is easy to use it.

knorth55:
I thought the interface can be similar to fcn semantic segmentation, and it is easy to understand 
How about publishing label image with instance ids?.
we can make mask with label image and labels.

wkentaro:
### Update 2018-04-05: add example which uses label images

```
mask_rcnn_instance_segmentation -> [label_cls] -> label_to_mask_image (select person class) -> apply_mask_image (remove instances other than person) -> label_image_decomposer (visualize)
                                -> [label_ins] -> ------------------------------------------|
```



I added an example of using label images that contains class and instance id.

wkentaro:
Kinetic build fails:

```
CMakeFiles/jsk_recognition_utils.dir/src/pcl_conversion_util.cpp.o: In function `jsk_recognition_utils::rangeImageToCvMat(pcl::RangeImage const&, cv::Mat&)':


pcl_conversion_util.cpp:(.text+0x19d): undefined reference to `pcl::visualization::FloatImageUtils::getColorForFloat(float, unsigned char&, unsigned char&, unsigned char&)'


pcl_conversion_util.cpp:(.text+0x1f1): undefined reference to `pcl::visualization::FloatImageUtils::getColorForFloat(float, unsigned char&, unsigned char&, unsigned char&)'


collect2: error: ld returned 1 exit status


make[2]: *** [/workspace/ros/ws_jsk_recognition/devel/.private/jsk_recognition_utils/lib/libjsk_recognition_utils.so] Error 1


make[1]: *** [CMakeFiles/jsk_recognition_utils.dir/all] Error 2


make: *** [all] Error 2


cd /workspace/ros/ws_jsk_recognition/build/jsk_recognition_utils; catkin build --get-env jsk_recognition_utils | catkin env -si  /usr/bin/make --jobserver-fds=6,7 -j; cd -




Failed     << jsk_recognition_utils:make             [ Exited with code 2 ]

Failed    <<< jsk_recognition_utils                  [ 36.0 seconds ]
```

PR #2300, opened 2018-06-19
[jsk_perception/openpose] Add resize image

iory:
When a large image is input, we cannot allocate openpose's neural network  on GPU memory.
So I added an option to resize for openpose nodelet.

### DIFF
--- a/jsk_perception/node_scripts/people_pose_estimation_2d.py
+++ b/jsk_perception/node_scripts/people_pose_estimation_2d.py
@@ -96,6 +96,8 @@ class PeoplePoseEstimation2D(ConnectionBasedTransport):
         self.pad_value = rospy.get_param('~pad_value', 128)
         self.thre1 = rospy.get_param('~thre1', 0.1)
         self.thre2 = rospy.get_param('~thre2', 0.05)
+        self.width = rospy.get_param('~width', None)
+        self.height = rospy.get_param('~height', None)
         self.gpu = rospy.get_param('~gpu', -1)  # -1 is cpu mode
         self.with_depth = rospy.get_param('~with_depth', False)
         self._load_model()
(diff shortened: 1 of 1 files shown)


furushchev:
Ah, I may just now understand the reason why you want to add resize on this node.
If the input image is resized before this node, output data in 2D image space (which is enabled when `~with_depth` is `False`) is affected and need to be re-scaled according to the original image size.

@iory 
Did you check on both `~with_depth` is `False` (hence, positional data in output is in coordinates of the original 2D image) and `~with_depth` is `True` (hence positional data is in 3D space)?
I asked because the current test case only checks if the output topic is published or not.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/test/people_pose_estimation_2d.test

If ok, I approve in my review.
(Optional) Improving the test code is also welcome. :+1:

jsk_perception/node_scripts/people_pose_estimation_2d.py
```diff
@@ -490,6 +503,8 @@ def _pose_estimate_chainer_backend(self, bgr_img):
                 bgr_img, joint_cands_indices, all_peaks, candidate)
         else:
             result_img = bgr_img
+        if not (self.width is None or self.height is None):
+            result_img = cv2.resize(result_img, (org_w, org_h))
 
         return result_img, self._extract_joint_position(joint_cands_indices, candidate)
```
wkentaro: @iory `self._extract_joint_position` is not affected by resizing the input for the NN?

k-okada:
waiting for testcode?

k-okada:
@iory : waiting for test code? or @iory self._extract_joint_position is not affected by resizing the input for the NN? ?

PR #2379, opened 2018-12-18
[jsk_pcl_ros_utils/cluster_point_indices_to_point_indices] Fixed document in case of index==-1

iory:


k-okada:
Is ‘concatenated cluster indices’ is ‘empty’ or not?

iory:
not empty

Issue #2380, opened 2018-12-23
attention_clipper does not work properly on Kinetic + PCL1.8.0


knorth55:
This is caused by `CropBox` in `pcl`.

On master branch
`roslaunch jsk_pcl_ros sample_attention_clipper.launch`

## Kinetic + PCL1.8.0 + Default configure
PCL build command
```bash
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
make -j $(nproc)
sudo make install
```

![attention_clipper_kinetic](https://user-images.githubusercontent.com/9300063/50385695-d2e1c180-071c-11e9-82b7-ee69c0c8751e.png)

```bash
rostopic echo /attention_clipper/output/point_indices
header: 
  seq: 1220
  stamp: 
    secs: 1486279001
    nsecs:  53697817
  frame_id: "right_hand_camera_rgb_optical_frame"
indices: []
---
header: 
  seq: 1221
  stamp: 
    secs: 1486279001
    nsecs:  87385647
  frame_id: "right_hand_camera_rgb_optical_frame"
indices: []
---
header: 
  seq: 1222
  stamp: 
    secs: 1486279001
    nsecs: 121011011
  frame_id: "right_hand_camera_rgb_optical_frame"
indices: []
---
```
## Kinetic + PCL1.8.0 + `-DPCL_ENABLE_SSE:BOOL=FALSE`:

PCL build command:
```bash
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release -DPCL_ENABLE_SSE:BOOL=FALSE ..
make -j $(nproc)
sudo make install
```

![attention_clipper_kinetic_sse](https://user-images.githubusercontent.com/9300063/50391530-b039c200-0789-11e9-817e-7b751fc15119.png)

## Indigo + PCL1.8.0:
![attention_clipper_indigo](https://user-images.githubusercontent.com/9300063/50385676-84342780-071c-11e9-89a0-5144cbc812c5.png)

## Kinetic + PCL 1.7.2
![attention_clipper_kinetic_1 7](https://user-images.githubusercontent.com/9300063/50390043-2f70cb00-0775-11e9-8a38-381b340b2126.png)

Probably?
https://github.com/PointCloudLibrary/pcl/issues/1623
https://github.com/PointCloudLibrary/pcl/pull/1917

knorth55:
I installed `PCL 1.8.1` and it doesn't solved.

[...]

PR #2381, opened 2018-12-24
set PCL_ENABLE_SSE:BOOL=FALSE in travis_before_script_pcl1.8.bash for kinetic build

knorth55:
Related to https://github.com/jsk-ros-pkg/jsk_recognition/issues/2380


iory:
I tested this option's behaviour in docker( #2382 ) and It works in case of `-DPCL_ENABLE_SSE:BOOL=FALSE` and it does not work in case of `-DPCL_ENABLE_SSE:BOOL=TRUE`.

PR #2388, opened 2019-01-21
add line to preserve rgb data of pointcloud

itohdak:
In `normal_estimation_omp_nodelet.launch`,  the output pointcloud doesn't take over the rgb data of the input pointcloud. This pull request will solve this issue.

I tested the result with `sample_door_handle_detector.bag` and checked if it works correctly.
Input pointcloud

The left is the previous output, and the right is the one with this pull request.


mmurooka:
+1
Could you double-check? > @iory

Issue #2378, opened 2018-12-13
There's typo in openni2_remote.launch and some other files.


ykawamura96:
`quater` should be `quarter` in those files below.
https://github.com/jsk-ros-pkg/jsk_recognition/search?utf8=%E2%9C%93&q=quater&type=
Although, it looks like difficult to fix these typos by simply replacing with correct one since many other files seem to be depending on these files.

PR #2391, opened 2019-01-29
[jsk_pcl_ros/launch/openni_launch, openni2_remote.launch] Fix typo quater -> quarter

iory:
This PR relates to https://github.com/jsk-ros-pkg/jsk_recognition/issues/2378
and modified typo quater to quarter.


k-okada:
There is already users who are using `quater`, if we want to fix this typo, we need transition plan.
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_sensors.xml#L36
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/kinect_head.launch#L99

PR #2329, opened 2018-09-12
OctomapServerContact sample with PR2

mmurooka:
Add sample of OctomapServerContact without gazebo.
Update README.
image

PR #2392, opened 2019-02-11
OctomapServerContact sample with PR2

k-okada:
rebased version of #2329


mmurooka:
I'm sorry that I did not do this.

PR #2408, opened 2019-03-14
ClusterPointIndicesDecomposer: suppress error if contains zero indices

furushchev:


jsk_pcl_ros/src/cluster_point_indices_decomposer_nodelet.cpp
```diff
@@ -416,6 +416,9 @@ namespace jsk_pcl_ros
    jsk_recognition_msgs::BoundingBox& bounding_box)
   {
     bounding_box.header = header;
+    if (segmented_cloud->points.size() == 0) {
+      return true;
```
k-okada: I am not sure, but should we output the warning message?

furushchev: Added.

PR #2414, opened 2019-03-14
fixes scope bug on point_pose_extraction

furushchev:
This PR fixes some bugs on `point_pose_extraction` node.
Since the `ObjectDetection_agg` topic here was just an alias of `ObjectDetection` and it was not aggregated, I created an aggregator. (see https://github.com/jsk-ros-pkg/jsk_common_msgs/pull/24 )


YutoUchimi:
For backward compatibility, I reverted the commit in which publishing of `ObjectDetection_agg` is removed.
If we need slower rate for _agg, let's make another PR.

PR #2417, opened 2019-03-25
Add qtbase5-dev as a package.xml dependency for jsk_recognition_utils.

clalancette:
Needed since CMakeLists.txt tries to depend on it.

This should fix the build errors seen on the Melodic buildfarm: http://build.ros.org/view/Mbin_ubhf_uBhf/job/Mbin_ubhf_uBhf__jsk_recognition_utils__ubuntu_bionic_armhf__binary/28/consoleFull


clalancette:
It looks like the CI builds failed with timeouts, but its not immediately clear to me what caused that.

k-okada:
thanks, done https://github.com/ros/rosdistro/pull/20730

PR #2426, opened 2019-05-07
Modify pcl version check for building with pcl-1.9

t141:
modified pcl version check for pcl-1.9


knorth55:
@t141 can you modify PR name to show what this PR is about?

iory:
@t141 
Did you commit by my name. 
Can you amend commit to change author?

PR #2430, opened 2019-05-27
[jsk_recognition_utils] Add label to bounding_box_publisher.py

taichiH:
Add label param to bounding_box_publisher.py


k-okada:
do we need to update https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/doc/jsk_recognition_utils/nodes/bounding_box_array_publisher.md ?

taichiH:
I think so.
I updated a document.

PR #2431, opened 2019-05-28
add negative rosparam in mask_image_filter

takayuki5168:
I added `negative` rosparam like [thi](https://jsk-recognition.readthedocs.io/en/latest/jsk_perception/nodes/apply_mask_image.html).

If someone thinks this PR as correct modification, I will also update docs.

PR #2441, opened 2019-06-29
add 'how to generate doc on local machine' in tips.rst

takayuki5168:
I updated doc with [your advice](https://github.com/jsk-ros-pkg/jsk_recognition/pull/2431#issuecomment-506929359)


takayuki5168:
@k-okada

PR #2440, opened 2019-06-24
Add sample, test and doc

YutoUchimi:
Close #1571

This PR adds sample, test and doc, and also fixes nodes.

### node

- [x] add_cluster_indices.py
- [x] image_16uc1_to_32fc1.py
- [x] polygon_array_publisher.py
- [x] polygon_array_to_polygon.py
- [x] pose_array_to_pose.py
- [x] rect_array_to_polygon_array.py
- [x] static_virtual_camera.py

### nodelet

- [x] LogPolar

### node

- [x] checkerboard_detector
- [x] checkerboard_calibration
    (o: sample, doc    x: test)
    (Cannot create test code)
- [x] objectdetection_transform_echo
- [x] objectdetection_tf_publisher.py

### node

- [x] object_array_publisher.py
- [x] people_pose_array_to_pose_array.py
- [x] plot_data_to_csv.py
- [x] save_mesh_server.py
    (o: test, doc    x: sample)
    (Cannot create sample code)

### node

- [x] apply_context_to_label_probability
- [x] binpack_rect_array.py
- ~detection_interface.l~ (This node requires `pr2eus`)
- [x] fast_rcnn.py
- [x] fcn_depth_prediction.py
    (often timeout, so disabled running test)
- [x] image_time_diff.py
- [x] kalman-filtered-objectdetection-marker.l
- [x] mask_rcnn_instance_segmentation.py
    (usually timeout, so disabled running test)
- [x] non_maximum_suppression.py
- ~objectdetection-marker.l~ (This node loads `detection_interface.l`, which requires `pr2eus`)
- ~people_mask_publisher.py~ (https://github.com/jsk-ros-pkg/jsk_recognition/pull/2347)
- [x] sliding_window_object_detector_trainer_node
    (o: sample, doc    x: test)
    (Cannot create test code)
- [x] pointit.py
- [x] solidity_rag_merge.py
- [x] unwrap_histogram_with_range_array.py

### nodelet

- [x] BlackHat
- [x] Closing
- [x] ColorHistogramLabelMatch
- ~ContourFinder~ (This node is deprecated.)
- [x] DilateMaskImage
- [x] ErodeMaskImage
- [x] FisheyeToPanorama
- [x] GaussianBlur
- [x] GrabCut
- [x] HSVDecomposer
- [x] LabDecomposer
- [x] MaskImageToROI
- [x] MorphologicalGradient
- [x] Opening
- [x] PolygonArrayColorHistogram
- [x] PolygonArrayColorLikelihood
[...]

PR #2442, opened 2019-07-04
fix generate_readme.py and update readme

knorth55:
- fix generate_readme.py
  - upstream script is modified
- update readme
  - debian package status was too old.


README.md
```diff
@@ -1,7 +1,7 @@
 <!--
```
YutoUchimi: Can you please generate README.md once more?
#2440 has been merged and a lot of documents with images are available now, so I think some new galleries will be created.

knorth55: there is no change because the max num of image is 15.

PR #2447, opened 2019-08-27
[jsk_recognition_utils] Fix ear clipping triangulation to support concave polygon

YutoUchimi:
Fix https://github.com/jsk-ros-pkg/jsk_visualization/issues/738

Current triangulation does not work correctly when concave polygon is passed.
The problem can be seen when the polygon is visualized in rviz (with `jsk_rviz_plugins`) like figures below.

This PR (based on https://github.com/PointCloudLibrary/pcl/pull/2743) will fix the problem, but there are a few differences not to use c++11.
- Do not use `cend()` for std::vector. Use `end()` instead.
- Do not use `emplace_back` for std::vector. Keep using `push_back`.

### Polygon edge (OK)
fix_ear_clipping1
fix_ear_clipping3

### Polygon surface before this PR (wrong)
__alpha = 1.0__
polygon_wrongly_filled2
polygon_wrongly_filled4

__alpha = 0.5__
polygon_wrongly_filled5
polygon_wrongly_filled6

### Polygon surface after this PR (OK)
__alpha = 1.0__
fix_ear_clipping2
fix_ear_clipping4

__alpha = 0.5__
polygon_wrongly_filled7
polygon_wrongly_filled8

### DIFF
--- a/jsk_recognition_utils/src/pcl/ear_clipping_patched.cpp
+++ b/jsk_recognition_utils/src/pcl/ear_clipping_patched.cpp
@@ -139,13 +139,13 @@ pcl::EarClippingPatched::isEar (int u, int v, int w, const std::vector<uint32_t>
   p_vu = p_u - p_v;
   p_vw = p_w - p_v;
 
-  // Avoid flat triangles and concave vertex
+  // 1: Avoid flat triangles and concave vertex
   Eigen::Vector3f cross = p_vu.cross(p_vw);
   if ((cross[2] > 0) || (cross.norm() < eps))
     return (false);
 
   Eigen::Vector3f p;
-  // Check if any other vertex is inside the triangle.
+  // 2: Check if any other vertex is inside the triangle.
   for (int k = 0; k < static_cast<int> (vertices.size ()); k++)
   {
     if ((k == u) || (k == v) || (k == w))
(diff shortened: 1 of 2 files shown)


jsk_recognition_utils/src/pcl/ear_clipping_patched.cpp
```diff
@@ -75,77 +75,56 @@ pcl::EarClippingPatched::triangulate (const Vertices& vertices, PolygonMesh& out
...
-    remaining_vertices = vertices.vertices;
-  else
-    for (int v = 0; v < n_vertices; v++)
+  std::vector<uint32_t> remaining_vertices = vertices.vertices;
+  size_t count = triangulate(remaining_vertices, output);
+
+  // if the input vertices order is anti-clockwise, it always left a
+  // convex polygon and start infinite loops, which means will left more
+  // than 3 points.
+  if(remaining_vertices.size() < 3)return;
```
knorth55: `if (remaining_vertices.size() < 3) return;`

jsk_recognition_utils/src/pcl/ear_clipping_patched.cpp
```diff
@@ -75,77 +75,56 @@ pcl::EarClippingPatched::triangulate (const Vertices& vertices, PolygonMesh& out
...
+  output.polygons.erase(output.polygons.end(), output.polygons.end() + count);
+  remaining_vertices.resize(n_vertices);
+  for (size_t v = 0; v < n_vertices; v++)
       remaining_vertices[v] = vertices.vertices[n_vertices - 1 - v];
+  triangulate(remaining_vertices, output);
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////////
+size_t
+pcl::EarClippingPatched::triangulate (std::vector<uint32_t>& vertices, PolygonMesh& output)
```
knorth55: I think it is better to name different function name from `triangulate` because it may cause misunderstanding.
how about `triangulate_step` ?

jsk_recognition_utils/src/pcl/ear_clipping_patched.cpp
YutoUchimi: Thank you for your review! Updated.

jsk_recognition_utils/src/pcl/ear_clipping_patched.cpp
YutoUchimi: Renamed: `triangulate` -> `triangulateClockwiseVertices`.

YutoUchimi:
I found a bug in `isEar` function.
According to https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf, "the line segment fromVi0 toVi2 lies completely inside the polygon", but there was no function to check that constraint.
I added it in b418fb1 .

The problem can be visualized by the simple code like below.
```xml
<launch>
  <node name="polygon_array_publisher"
        pkg="jsk_recognition_utils" type="polygon_array_publisher.py">
    <rosparam>
      publish_rate: 10.0
      frame_id: map
      polygons:
        - points:
            - [1, 1, 0]
            - [1, 4, 0]
            - [0, 3, 0]
            - [-1, 4, 0]
          label: 1
          likelihood: 1.0
        - points:
            - [1, -4, 0]
            - [0, -3, 0]
            - [-1, -4, 0]
            - [-1, -1, 0]
          label: 2
          likelihood: 1.0
        - points:
            - [-1, 1, 0]
            - [-4, 1, 0]
            - [-3, 0, 0]
            - [-4, -1, 0]
            - [-1, -1, 0]
          label: 3
          likelihood: 1.0
    </rosparam>
  </node>
</launch
```

#### before b418fb1 
fix_ear_clipping10

#### after b418fb1 
fix_ear_clipping11

PR #2452, opened 2019-10-02
Fix tile_image.py for Python3

wkentaro:
## Why?

```
process[data_collection_server-5]: started with pid [28484]
  File "/home/wkentaro/xxxxxxxxxxxx/src/jsk-ros-pkg/jsk_recognition/jsk_perception/node_scripts/tile_image.py", line 81
    async = message_filters.ApproximateTimeSynchronizer(
          ^
SyntaxError: invalid syntax
```


k-okada:
Same as https://github.com/jsk-ros-pkg/jsk_recognition/pull/2454

`aync`: https://docs.python.org/ja/3/library/asyncio.html

PR #2454, opened 2019-10-11
Fix label_image_decomposer.py for Python3

wkentaro:


knorth55:
`aync`: https://docs.python.org/ja/3/library/asyncio.html

PR #2457, opened 2019-10-26
fix xtion2 installation document

sktometometo:
fix xtion2 installation document because SDK files that can be downloaded from ASUS web sitte had changed.


k-okada:
closed via https://github.com/jsk-ros-pkg/jsk_recognition/pull/2458

PR #2459, opened 2019-10-26
add Structure core Installatin Document

sktometometo:
- add a Structure core Installation document

### DIFF
--- a/doc/install_structure_core.rst
+++ b/doc/install_structure_core.rst
@@ -1,5 +1,5 @@
 Install Structure core Camera
-====================
+=============================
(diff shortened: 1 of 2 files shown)


doc/install_structure_core.rst
```diff
@@ -0,0 +1,55 @@
+Install Structure core Camera
+====================
```
knorth55: put same length `=` as the first line.

sktometometo:
Updated index.rst and fixed some syntax mistake in install_structure_core.rst

PR #2461, opened 2019-10-30
add the on-off function of using use_pca in dynamic reconfigure

poyotamu000:
When I created the simple picking program, I added on-off function of using use_pca in dynamic reconfigure. I think it is more convenient.


YutoUchimi:
Thank you for your contribution!
This feature is convenient.

How about remove  `pnh_->param("use_pca", use_pca_, false);` ?
The parameter is initialized twice by this line.

PR #2465, opened 2019-11-05
[jsk_perception/people_pose_estimation_2d.py][jsk_perception/people_mask_publisher.py] Fix edge case bug

itohdak:
In Numpy, since ```a[0:i]``` with negative ```i``` returns the values from index ```0``` to index ```n+i``` (assuming ```n``` is the array length), we got unexpected cropped images in some edge cases.

Although ```top < bottom``` are guaranteed, if we get some values such that ```top < bottom < 0```, the following ```cropped``` becomes an array whose size is ```(imh+bottom, *)```, which is expected to be ```(0, *)```. The same can be said with ```left``` and ```right```.

```
cropped = img[max(0, top):min(imh, bottom), max(0, left):min(imw, right)]
```

I fixed it by taking ```max``` with ```0``` also in the second value (end index) in slice.

**Sample videos:**
result
**left up:** people_pose_estimation_2d/output,
**right up:** people_mask_publisher/debug/output,
**left bottom:** people_mask_publisher/output w/o this PR,
**right bottom:** people_mask_publisher/output w/ this PR


YutoUchimi:
@itohdak 
Can you share a sample image which shows before/after this PR?

knorth55:
Why does negative number appear in bounding boxes?

itohdak:
@YutoUchimi 
Okay, I added an sample video.

itohdak:
@knorth55 
In `people_mask_publisher.py`, the hand center position is decided by calculating the externally dividing point with parameter `hand_ratio`, so even if the wrist and elbow positions are inside the image range, the estimated hand center position (or the edge of the cropped area) can go out of the range.

https://github.com/jsk-ros-pkg/jsk_recognition/blob/e9172e85b281ff0899bc5d429b2cdd56e8e455dd/jsk_perception/node_scripts/people_mask_publisher.py#L123-L137

YutoUchimi:
@itohdak 
Thanks! I understood.

knorth55:
@k-okada travis passed! can you check and merge this PR?

PR #2463, opened 2019-11-03
[jsk_pcl_ros] Add multi euclidean clustering

iory:
Add multi euclidean clustering as a `jsk_pcl/EuclideanClustering`'s option.
This PR enables processing `jsk_recognition_msgs/ClusterIndices`'s input at the same time.

The results of multi euclidean clustering as follows.
with multi euclidean clustering
https://drive.google.com/file/d/1ZMKwmq1TPtEdJ8mgWrSWlxiP58xU8QOu/view?usp=sharing

without multi euclidean clustering
https://drive.google.com/open?id=1vcRwDHxEnc2NbWidBeiEZau-nZGl2MXi

Those videos are in the same situation.

PR #2468, opened 2019-11-05
Add cluster point indices label filter

Affonso-Gui:
Talking to @YutoUchimi and @knorth55 we realized that we do not actually have any good methods to extract cluster_points from `mask_rcnn` with matching labels  _e.g._ getting only pet bottle objects from 73b2_kitchen recognition node. So we decided to implement a node to do that: `ClusterPointIndicesLabelFilter`.

Other approaches such as using [LabelToMaskImage](https://jsk-recognition.readthedocs.io/en/latest/jsk_perception/nodes/label_to_mask_image.html) are not only redundant, but also many times only cover the first label instance. Subscribing and filtering in eus is also not safe regarding synchronization.

The usage / design of this node is questionable, so am open to discussions / suggestions.

Specially regarding cases where we want to get multiple types of objects, was also thinking it might be a good idea to set a namespace and a label list and have filtered results under that namespace (using the label name). Something like:
id_value: [7,9,13]
namespace: 'test'
`test/plastic_bottle`
`test/can`
`test/table`

@iory


Affonso-Gui:
Btw regarding sample / test code, was thinking on making use of the hsr bag file on https://github.com/jsk-ros-pkg/jsk_recognition/pull/2463 , but for that we would need to have it on the `jsk_pcl_ros_utils` repo. 

@iory @YutoUchimi

iory:
I see. Should I mv the bag file to `jsk_pcl_ros_utils` on #2463 or make another PR?

Affonso-Gui:
@iory If you can move the bag file (and the play_rosbag xml) to the jsk_pcl_ros_utils in the same PR that would be appreciated!

iory:
OK. I moved the bag file to `jsk_pcl_ros_utils` https://github.com/jsk-ros-pkg/jsk_recognition/pull/2463/commits/dbe9239d9c3d725fd5e242d0edd7759aa97872dd

Affonso-Gui:
Added sample and test code!

Affonso-Gui:
NOTE: this requires #2463 to be merged first

PR #2470, opened 2019-11-12
Parameterize frames, transformation and interpolation in virtual_camera_mono

YutoUchimi:
- Use latest available transform for lookupTransform
- Parameterize frame_id, transformation between target frame and virtual camera, and interpolation method.
- Add sample, test and doc.


YutoUchimi:
Required in https://github.com/jsk-ros-pkg/jsk_demos/pull/1276

PR #2472, opened 2019-11-17
[jsk_pcl_ros] Add nearest plane index label to cluster_point_indices_decomposer

iory:
Currently, each box label of output boxes of cluster_point_indices_descomposer is always 0.
With this PR, If `~align_boxes` and `~align_boxes_with_plane` are `True`, each 
box(`jsk_recognition_msgs/BoundingBox`)'s label indicates nearest plane index.

The visualized results as follows:
With this PR
Selection_002

Without this PR
Selection_001

### DIFF
--- a/doc/jsk_pcl_ros/nodes/cluster_point_indices_decomposer.md
+++ b/doc/jsk_pcl_ros/nodes/cluster_point_indices_decomposer.md
@@ -27,6 +27,8 @@ It also publishes tf of centroids of each cluster and oriented bounding box of t
 
    Array of oriented bounding box for each segmented cluster.
 
+   If `~align_boxes` and `~align_boxes_with_plane` are `True`, each box(`jsk_recognition_msgs/BoundingBox`)'s label indicates nearest plane index.
+
 * `~label` (`sensor_msgs/Image`):
 
    Label image for each cluster point indices.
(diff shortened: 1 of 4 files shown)


knorth55:
I think it is confusing to add such index in `bounding_box.label` is used as `bounding_box` instance id when sort is enabled.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/021d50727115e0ed7715cdb7cbe4021d6a53aece/jsk_pcl_ros/src/cluster_point_indices_decomposer_nodelet.cpp#L696
@YutoUchimi What do you think?

YutoUchimi:
I agree with @knorth55, but this feature seems useful.
How about adding a new parameter `~fill_with_nearest_plane_index` or something like that in order to keep backward compatibility?

iory:
I add `~fill_bba_label_with_nearest_plane_index` parameter to keep backward compatibility.
Please review the changes. ( `~fill_bba_label_with_nearest_plane_index` is `False` in default)

The results as follows:
`~fill_bba_label_with_nearest_plane_index` is True.

jsk_startup rviz- - RViz_005

`~fill_bba_label_with_nearest_plane_index` is False.

jsk_startup rviz- - RViz_006

YutoUchimi:
@iory 
Is this PR still WIP? If no, could you remove it from the title of this PR?

iory:
Oh, I forgot it. Remove it, thanks!

knorth55:
what is `bba`?

iory:
`bba` stands for  `bounding box array`

knorth55:
the abbreviation is a bit difficult...
how about `~fill_boxes_with_nearest_plane_index`?

iory:
OK. Renamed bba to boxes. Please check it.

PR #2476, opened 2019-11-28
[jsk_perception] apply_mask_image publish roi

kosuke55:
When camera_info is given, publish camera_info with roi of mask image.


knorth55:
BTW, how can you use this camera info?
which node requires this roi values in camera info?
can you paste screen shot of your sample, too?

jsk_perception/src/apply_mask_image.cpp
```diff
@@ -113,12 +125,22 @@ namespace jsk_perception
       cv::bitwise_not(mask, mask);
     }
 
+    cv::Rect region = jsk_recognition_utils::boundingRectOfMaskImage(mask);
+    if (camera_info_){
+      sensor_msgs::CameraInfo roi(*camera_info_);
```
knorth55: change name from `roi` to `camera_info`.

jsk_perception/src/apply_mask_image.cpp
```diff
@@ -113,12 +125,22 @@ namespace jsk_perception
...
+      roi.roi.x_offset = region.x;
+      roi.roi.y_offset = region.y;
+      roi.roi.width = region.width;
+      roi.roi.height = region.height;
+      roi.roi.do_rectify = true;
```
knorth55: This is just a question, but do we need to set this as true and can we set this as true?
```
# True if a distinct rectified ROI should be calculated from the "raw"
# ROI in this message. Typically this should be False if the full image
# is captured (ROI not used), and True if a subwindow is captured (ROI
# used).
```

kosuke55:
@knorth55 
> this is just a question but can we set this variable as true in this case?

I set true without thinking too much. I think It is better to give true or false with rosparam whether to use rectified image as input. How about it?

> change name from roi to camera_info.

OK, I will change it.
For reference, attention_clipper also used the name roi.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/81b136d28dfa6f036e55125cb3e2861093123cce/jsk_pcl_ros/src/attention_clipper_nodelet.cpp#L361

kosuke55:
@knorth55 
In my use case, reflect the detection result in the clipped image to the original image. 
I think this is useful when the original image is large and processing is heavy, or when the object is too small to be detected.

clip
Detection result in clipped image

raw
Reflect in the original image

jsk_perception/src/apply_mask_image.cpp
kosuke55: OK, I will change it.
For reference, attention_clipper also used the name roi.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/81b136d28dfa6f036e55125cb3e2861093123cce/jsk_pcl_ros/src/attention_clipper_nodelet.cpp#L361

jsk_perception/src/apply_mask_image.cpp
kosuke55: I set true without thinking too much. I think It is better to give true or false with rosparam whether to use rectified image as input. How about it?

jsk_perception/src/apply_mask_image.cpp
knorth55: we should make code better.

knorth55 (approved):

iory (approved):

k-okada (changes requested):
thanks for contribution. could you add documentation for this? c.f.https://github.com/jsk-ros-pkg/jsk_recognition/pull/2535/files#diff-9cba1daceb335f6f97bf7980d0d51ec0

kosuke55:
@k-okada 
I added the documentation. Please check it.  
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2476/files#diff-96d5060a4501ac15a1ff116055e08cba

k-okada (approved):

PR #2480, opened 2019-12-14
set chainer version less than 7.0.0

knorth55:
chainer v7.0.0 drops python2.7 support.

I don't know why these people love breaking dependencies... :(

PR #2485, opened 2019-12-21
set chainer version less than 7.0.0

k-okada:
rewrited version of #2480 with https://github.com/ros-infrastructure/rosdep/pull/694 (support version_lt for pip)

- use https://github.com/ros-infrastructure/rosdep/pull/694 to respect  version_lt for python pip, some package requries python3
- set chainer version less than 7.0.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_script:
   - if [[ "$ROS_DISTRO" == "melodic" && "$DOCKER_IMAGE_JENKINS" == "" ]]; then export export DOCKER_IMAGE_JENKINS='ros-ubuntu:18.04-pcl'; fi
 script:
   # use https://github.com/ros-infrastructure/rosdep/pull/694 to respect version_lt for python pip, some package requries python3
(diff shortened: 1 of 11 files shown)


knorth55:
https://github.com/chainer/chainer/pull/8517 is merged and chainer v7.1.0 is released.

pazeshun:
But the problem still remains:
https://github.com/chainer/chainer/pull/8517#issuecomment-576128957

k-okada:
install following versions manually, because
1) https://github.com/ros-infrastructure/rosdep/pull/694 does not work as expected on all cases..
2) whith system numpy, the fcn hungs (can be reproduced on docker machine@jenkins, but not reproducable on local environment....)
```
  - export BEFORE_SCRIPT="sudo pip install numpy==1.16.6; $BEFORE_SCRIPT"
  - export BEFORE_SCRIPT="sudo pip install fcn chainercv chainer==6.7.0; $BEFORE_SCRIPT"
```

PR #2497, opened 2020-05-14
[doc] add install pcl from source doc

knorth55:
add pcl build doc for sse build error and kinfu

cc. @Kanazawanaoaki

### DIFF
--- a/doc/install_pcl_from_source.md
+++ b/doc/install_pcl_from_source.md
@@ -1,0 +1,37 @@
+# Install PCL from source
+
+## Install PCL from source without SSE
+
+This is needed for [AttentionClipper](https://github.com/jsk-ros-pkg/jsk_recognition/issues/2380).
+
+```bash
+cd
+wget https://github.com/PointCloudLibrary/pcl/archive/pcl-1.8.1.tar.gz
+tar xvzf pcl-1.8.1.tar.gz
+cd pcl-pcl-1.8.1
+mkdir build
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release -DPCL_ENABLE_SSE:BOOL=FALSE
+make -j 4
+sudo make install
+```
+
+And rebuild all PCL related packges from source
+
+## Install PCL from source with CUDA
+
+This is needed for Kinfu.
+
+```bash
+cd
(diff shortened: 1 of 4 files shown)


doc/install_pcl_from_source.md
```diff
@@ -0,0 +1,37 @@
+# Install PCL from source
+
+## Install PCL from source without SSE
+
+This is needed for [AttentionClipper](https://github.com/jsk-ros-pkg/jsk_recognition/issues/2380).
```
pazeshun, line 5: I prefer `This is sometimes needed for AttentionClipper`, because this installation is not always required (and perhaps we won't need this procedure on default environment in the future).

doc/install_pcl_from_source.md
```diff
@@ -0,0 +1,37 @@
+# Install PCL from source
+
```
pazeshun, line 2: Can the following two settings be applied at the same time?
If so, I think it is better to write `The following settings can be applied at the same time by combining cmake options` here.

doc/jsk_pcl_ros/nodes/attention_clipper.md
```diff
@@ -8,6 +8,14 @@ retrieves `sensor_msgs/PointCloud2` and publish `pcl_msgs/PointIndices`.
 You can specify the pose and size of the interest bounding box and jsk\_pcl/AttentionClipper returns ROI
 to see the object.
 
+## Note
+
+AttentionClipper does not work properly on Kinetic + PCL 1.8.0, and you can avoid the problem with SSE disabled PCL.
```
pazeshun, line 13: Also on melodic?

knorth55: I didn't check it on melodic.
Did you check it on melodic?

doc/install_pcl_from_source.md
knorth55: thx, I updated.

doc/install_pcl_from_source.md
knorth55: thx, I updated.

doc/jsk_pcl_ros/nodes/attention_clipper.md
pazeshun: Oops, sorry, I was confused with kinfu.
Please forget my comment.

PR #2500, opened 2020-05-18
fix print syntax in train_ssd.py

mqcmd196:
Because of `jsk_perception/scripts/train_ssd.py` 's syntax error in l165, the error message couldn't be shown correctly. This python2 script includes `print_function` from `__future__`, so I think my style is correct. I confirmed work correctly.

Before
```bash
$ rosrun jsk_perception train_ssd.py dataset_voc/
  File "/home/yoshiki/research_ws/src/jsk_recognition/jsk_perception/scripts/train_ssd.py", line 165
    print 'unsuppported dataset type'
                                    ^
SyntaxError: invalid syntax
```

After
```bash
$ rosrun jsk_perception train_ssd.py dataset_voc/
Please install CuPy

    sudo pip install cupy-cuda[your cuda version]
i.e.
    sudo pip install cupy-cuda91
```

Thanks.

### DIFF
--- a/jsk_perception/scripts/train_ssd.py
+++ b/jsk_perception/scripts/train_ssd.py
@@ -162,7 +162,7 @@ def main():
     elif (args.dataset_type == 'bbox'):
         train_dataset = BboxDetectionDataset(args.train_dataset_dir)
     else:
-        print 'unsuppported dataset type'
+        print('unsuppported dataset type')
         return
 
     fg_label_names = train_dataset.fg_class_names


knorth55:
@mqcmd196 good PR!
can you change title to describe the problem more like "fix print syntax in train_ssd.py"?
also, can you check which python you are using?
you can check it with `which python`, and `/usr/bin/python` is the right one.
if not, it would cause serious error when you are running python in ROS.

mqcmd196:
Thanks for your reviewing!

I checked my python environment.
`which python` shows `usr/bin/python`.
`python -V` shows `Python 2.7.17`

k-okada:
```
$ python -V
Python 2.7.17
$ python -c "print 'hoge'"
hoge
$ python -c "from __future__ import print_function; print 'hoge'"
  File "<string>", line 1
    from __future__ import print_function; print 'hoge'
                                                      ^
SyntaxError: invalid syntax
```

The error is intdocued in https://github.com/jsk-ros-pkg/jsk_recognition/commit/94c21353b9e369fc4d9167e47cd84119a96d92b2 and send another PR #2502

PR #2507, opened 2020-05-26
Fix for  noetic / 20.04

k-okada:
see http://wiki.ros.org/noetic/Migration for migration


knorth55:
@k-okada 
I have a question about `catkin_install_python`.
Don't we need to install all `node_scripts` in `catkin_install_python` for shebang issue?
In my understanding, we need to change to use `catkin_install_python` for shebang rewriting for python3 as https://github.com/ros-visualization/rqt_graph/pull/43

k-okada:
yes, but I do not want to assign this task to you.

--
◉ Kei Okada



2020年8月28日(金) 15:17 Shingo Kitagawa <notifications@github.com>:

knorth55:
@k-okada Thank you. I just want to know it.

PR #2511, opened 2020-06-09
[jsk_recognition_utils] add rect_array_to_cluster_point_indices.py

knorth55:
add `rect_array_to_cluster_point_indices.py`
this node converts `RectArray` to `ClusterPointIndices`, which is useful for object detection nodes.

rects_to_cpi

it is related to #2467

PR #2512, opened 2020-06-10
[Merge after #2511][jsk_pcl_ros] add MultiEuclideanClustering

knorth55:
Originally from @taichiH and @iory in https://github.com/taichiH/neatness_estimator
Merge after #2511 

add `MultiEuclideanClustering` nodelet.
this is useful for object detection and instance segmentation.

multi_euclidean_clustering_doc

cc. @taichiH @iory


iory:
Is this PR a duplication of https://github.com/jsk-ros-pkg/jsk_recognition/pull/2463 ?

knorth55:
>Is this PR a duplication of #2463 ?

That's true.
Anyway, we can add both, I think.

PR #2524, opened 2020-09-07
Always include cython3 when searching

PeterMitrano:
When using jsk recognition on 20.04, where there is no python or cython 2, it failed to find cython3. This fixes that issue.


k-okada:
@PeterMitrano thank you for contribution. closing in faver of #2531

PR #2544, opened 2020-11-12
[doc] Add insta360 air camera doc

knorth55:
add insta360 air camera

insta360_air_sample

cc. @708yamaguchi


708yamaguchi (commented):
Thank you.

Why do you think we need `jsk_perception/launch/libuvc_camera.launch`?
I think `sample_insta360_air.launch` is enough.

708yamaguchi:
We may need `libuvc_camera` to package.xml.

knorth55:
libuvc_camera launch is also used in #2545 , that is why i added it.
libuvc_camera is needed to run,
but it is just sample so i didn't add it.
i dont know how hard the release is, but i think we should not add a lot of run dependencies for sample programs

708yamaguchi (approved):
OK, thank you for your explanation.

knorth55:
@k-okada Travis passed, can you please merge this PR?

708yamaguchi:
FYI:
The mjpeg format of libuvc_camera is not available in indigo and fetch15 cannot use this launch file.
The following comment is helpful.
https://github.com/ros-drivers/libuvc_ros/issues/37#issuecomment-341284169

Instead, I was able to view  the insta360 mjpeg images by using usb_cam.

708yamaguchi:
@knorth55 @YoheiKakiuchi how can I get panorama view image  from 360 fisheye camera?

knorth55:
@708yamaguchi what do you mean by panorama view image?
if you mean the rectified image, you need to calibrate the camera.

708yamaguchi:
I mean panorama view image like below.
http://rraj.rsj-web.org/atcl/749

If calibration is needed, do you know a good method to calibrate the camera?

knorth55:
I think it is better to calibrate the camera as normal camera, split into several images and recognize objects and human separately.

FYI:
fisheye calibration: https://github.com/ros-perception/image_pipeline/pull/440
fisheye camera model: https://github.com/ros-perception/vision_opencv/pull/358
fisheye image_proc: https://github.com/ros-perception/image_pipeline/pull/441

YoheiKakiuchi:
Is there a recognition technique that uses original fisheye image ?
Opencv calibration is not so easy with a checkerboard, which is usually done with a normal lens, because the distortion is so large.

The opencv calibration just transform from a fisheye model to a pinhole model.
So you'll need to use narrow angle of view, or more than two images. 

I think the panorama view is using N images which are connected in a nice way,
but I can't find a paper or an implementation.

knorth55:
> Is there a recognition technique that uses original fisheye image ?
Opencv calibration is not so easy with a checkerboard, which is usually done with a normal lens, because the distortion is so large.

Which version of opencv have you done with?
From opencv3, fisheye camera model and calibration is introduce, so I think it is not so hard as it was (I hope).
`image_pipeline` camera calibration now supports fisheye calibration, so we can try it.
(however, we still don't have the image_proc for fisheye camera model)

https://stackoverflow.com/questions/50857278/raspicam-fisheye-calibration-with-opencv
https://docs.opencv.org/3.0.0/db/d58/group__calib3d__fisheye.html

>I think the panorama view is using N images which are connected in a nice way,
but I can't find a paper or an implementation.

If you want to make panorama view with no calibration, we need to spin the camera in z-axis, capture images and connect it, in my understanding.
If there is any technique to make a panorama from no calibrated fisheye camera, I really want to know it.

knorth55:
@708yamaguchi BTW, can you also add launch file for usb_cam and add args `use_usb_cam` in `sample_insta360.launch`?

708yamaguchi:
@YoheiKakiuchi @knorth55 
Thank you very much for your information.
I understand that I need to find good method by myself if I want to create panorama view or to do fisheye rectification on ROS...

@knorth55 
I sent PR to you.
Please merge this if there is no problem.
https://github.com/knorth55/jsk_recognition/pull/3

knorth55:
@k-okada Travis passed! Can you merge this PR?

PR #2545, opened 2020-11-12
[doc] Add elp camera doc

knorth55:
add elp 4k usb camera doc

elp_4k_usb_sample

cc. @708yamaguchi


708yamaguchi:
I was able to use this camera!

How about adding `<node if="$(arg gui)" pkg="image_view" ...>` for this sample launch file and https://github.com/jsk-ros-pkg/jsk_recognition/pull/2544?

knorth55:
thanks, I add `gui` args in both #2544 and #2545

708yamaguchi (approved):

PR #2564, opened 2021-01-11
[jsk_pcl_ros] add z range option to heightmap converter

sktometometo:
- add z range option to heightmap converter
- add descrptions for dynamics reconfigure parameters of heightmap converter


k-okada (changes requested):
could you update doc too? thanks. https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/doc/jsk_pcl_ros/nodes/heightmap_converter.md#L47

PR #2565, opened 2021-01-13
[jsk_recognition_utils] use try to import chainer depend modules

knorth55:
use `try` to import `chainer` depending modules.
this PR solves https://github.com/jsk-ros-pkg/jsk_visualization/issues/800

cc. @Naoki-Hiraoka

### DIFF
--- a/jsk_recognition_utils/python/jsk_recognition_utils/__init__.py
+++ b/jsk_recognition_utils/python/jsk_recognition_utils/__init__.py
@@ -1,5 +1,7 @@
 from __future__ import print_function
 
+import sys
+
 from jsk_recognition_utils import color
 from jsk_recognition_utils import conversations
 from jsk_recognition_utils import feature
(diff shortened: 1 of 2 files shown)


jsk_recognition_utils/python/jsk_recognition_utils/__init__.py
```diff
@@ -1,12 +1,15 @@
...
 from jsk_recognition_utils import feature
 from jsk_recognition_utils import mask
 from jsk_recognition_utils import visualize
 from jsk_recognition_utils import geometry
 
+try:
+    from jsk_recognition_utils import chainermodels  # NOQA
+    from jsk_recognition_utils import datasets  # NOQA
+except ImportError:
+    print('Please install chainer<7.0.0 to import chainermodels')
```
pazeshun, line 12: `to import chainermodels` -> `to import chainermodels and datasets`?
(Because we have to notice this message when we face error about `datasets` module as well as `chainermodels` module.)

pazeshun (changes requested):

jsk_recognition_utils/python/jsk_recognition_utils/__init__.py
knorth55: that's true. I will fix it.

Naoki-Hiraoka:
In additon to `chainer`, I found that `jsk_recognition_utils/chainermodels` also depends on
`fcn`

pazeshun:
I also found `jsk_recognition_utils/datasets` depends on `chainercv`:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/2c1400e15d79b4f5f9fc7e6784a58bb277ebd8e3/jsk_recognition_utils/python/jsk_recognition_utils/datasets/bbox_detection.py#L6
I think it is troublesome to list all required packages in the printing message, so how about `Please install chainer<7.0.0 and other required packages to import jsk_recognition_utils.chainermodels and datasets`?

knorth55:
@pazeshun I update this branch. I found that `chainer`, `chainercv` and `fcn` are `chainer` depending packages, so I check these 3 packages. How about this?

pazeshun:
@knorth55 Oh, thank you for your hard work!
After this change, we have to change `jsk_recognition_utils/__init__.py` as well as `package.xml` when new dependency is added to `chainermodels` or `datasets`.
Is this OK for you?

knorth55:
@pazeshun yes, probably it is ok.
also all chainer problems will be solved when all robots use `noetic` + `python3`...

pazeshun (approved):

PR #2568, opened 2021-01-15
[jsk_pcl_ros] remove wrong launch_manager in tabletop_object_detector.launch

knorth55:
remove `launch_manager` in `OrganizedMultiPlaneSegmentation`
this is wrong.

cc. @tohirose


knorth55:
@k-okada can you review and merge this PR? the launch was pretty buggy with this wrong `launch_manager` arg.

knorth55:
@k-okada can you merge this PR?

k-okada (commented):
do we remove 
```
  <arg name="launch_manager" default="true" />
```
too?

knorth55:
@k-okada no, we need `launch_manager` argument to disable nodelet manager.
current master branch disables `OrganizedMultiPlaneSegmentation` with `launch_manager` argument, which is not manager.

https://github.com/knorth55/jsk_recognition/blob/fix-tabletop-object-detector/jsk_pcl_ros/sample/tabletop_object_detector.launch#L26-L29

PR #2571, opened 2021-01-30
[jsk_pcl_ros] set larger min_size limit for organized multiplane segmentation

knorth55:
this PR change `min_size` limit of `OrganizedMulitPlaneSegmentation` to int max limit.
current number `10000` is too small for recent high resolution cameras.

cc. @tohirose

### DIFF
--- a/jsk_pcl_ros/cfg/OrganizedMultiPlaneSegmentation.cfg
+++ b/jsk_pcl_ros/cfg/OrganizedMultiPlaneSegmentation.cfg
@@ -8,7 +8,7 @@ from dynamic_reconfigure.parameter_generator_catkin import *;
 from math import pi
 
 gen = ParameterGenerator ()
-gen.add("min_size", int_t, 0, "the minimum number of the points of each cluster", 2000, 0, 10000)
+gen.add("min_size", int_t, 0, "the minimum number of the points of each cluster", 2000, 0, 2147483647)
 gen.add("distance_threshold", double_t, 0, "distance threshold of organized plane segmentation", 0.01, 0, 1.0)
 gen.add("angular_threshold", double_t, 0, "angular threshold of organized plane segmentation", 0.05, 0, 0.1)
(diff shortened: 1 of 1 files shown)


jsk_pcl_ros/cfg/OrganizedMultiPlaneSegmentation.cfg
```diff
@@ -8,7 +8,7 @@ from dynamic_reconfigure.parameter_generator_catkin import *;
 from math import pi
 
 gen = ParameterGenerator ()
-gen.add("min_size", int_t, 0, "the minimum number of the points of each cluster", 2000, 0, 10000)
+gen.add("min_size", int_t, 0, "the minimum number of the points of each cluster", 2000, 0, 2147483647)
```
k-okada, line 11: how did yuo caculate 2147483647 ? that is ~ (* 15731 15731). larger than 640 x 480, so what is the definition of 'recent high resolution cameras.'?

Suggested change is to add more commen. Something like 
"the minimum number of the points of each cluster (for high-reso cameras i.e. 1280x740, expected number of each cluster is 10% of total point ....." )

knorth55: this numeber is the max limit of int_t. i se this value because, for the firther future, we dont need to change this parameter any more because it is the limit.

knorth55: however we can change the value to 4K or 8K for the convenience.

knorth55: this PR only changes the max value of `min_size` and not changing the default value.

k-okada: sorry, I thought you changed the default value.. Sorry for that.

PR #2583, opened 2021-03-24
add aws_auto_checkin_app

k-okada:
See https://aws.amazon.com/jp/builders-flash/202004/auto-checkin-app/ for more info and https://github.com/jsk-ros-pkg/jsk_demos/pull/1325 for original codes


k-okada:
@iory, could you send me your face image file? https://github.com/k-okada/jsk_recognition/blob/add_auto_checking/jsk_perception/sample/sample_aws_auto_checkin_app.launch uses sample/data/2017-06-20-12-00-00_people_images_in_lab.bag whcih record your motion.
we'll upload your file to S3 and generate recognition database. Your face image itself will not published. see https://aws.amazon.com/jp/builders-flash/202004/auto-checkin-app/ for how it works.

iory:
That's interesting!
I sent a face image to your email address.
Please check it.

PR #2585, opened 2021-04-09
Fix travis 

knorth55:
- fix decorator version
  - decorator >= 5.0.0 is only for python3 but released for python2
  - https://github.com/micheles/decorator/pull/100


knorth55:
travis fails because of this change.
 we need to wait for new release of jsk travis or change to use master branch.

https://github.com/jsk-ros-pkg/jsk_travis/issues/417

knorth55:
Travis passed!

PR #2589, opened 2021-04-22
Implementation TrTr  visual tracker

tongtybj:
I add my original visual tracker call TrTr (tracker with Transformer) into jsk_perception.

For implantation and tutorial, please check [doc/jsk_perception/nodes/trtr_tracking.md](https://github.com/tongtybj/jsk_recognition/blob/936a2748ab161a1c085bd70d2c22184012c1a995/doc/jsk_perception/nodes/trtr_tracking.md).

The original implementation is based on [pytorch](https://github.com/tongtybj/TrTr), of which the inference time is not fast enough.
Then I use TensorRT to support FP16 for fast inference, and also use ONNX RunTIme library to support inference in CPU (unfortunately slow). 

The biggest problem of this RP is that TensorRT requires CUDA >= 10.2, which is incompatible  with other Chainer based node in jsk_perception.


tongtybj:
@knorth55  @sktometometo 

Why can't I assign reviewer when I create this PR...?

PR #2608, opened 2021-07-07
fix typo in github workflow branch name

knorth55:
I fix branch name from `main` -> `master` in github workflow branch name


k-okada (approved):
thanks!

PR #2623, opened 2021-09-02
[jsk_perception] Add approximate sync true in sample mask rcnn instance segmentation

poyotamu000:
…rosparam approximate_sync is True

I used this sample to try out the chainer environment for jsk_recognition. 
However, on my laptop, the viewer did not launch without changing the rosparam, `approximate_sync : True`. 
On a desktop PC, it started without changing the rosparam. 
I think it depends on the performance of  PC, but since the viewer doesn't start up without any error, I created a pull request.


jsk_perception/sample/sample_mask_rcnn_instance_segmentation.launch
```diff
@@ -60,6 +60,8 @@
     <remap from="~input/label" to="apply_mask_image/output" />
     <rosparam subst_value="true">
       bg_label: -1
+      approximate_sync: True
+      slop: 1.0
```
knorth55, line 64:
```suggestion
      approximate_sync: $(arg approximate_sync)
      slop: $(arg slop)
```

knorth55: and add these at the top
```
<arg name="approximate_sync" default="true" />
<arg name="slop" default="1.0" />
```

knorth55 (changes requested):

jsk_perception/sample/sample_mask_rcnn_instance_segmentation.launch
poyotamu000: Thank you very much. The code has been updated to reflect your suggestions.

jsk_perception/sample/sample_mask_rcnn_instance_segmentation.launch
```diff
@@ -60,6 +63,8 @@
     <remap from="~input/label" to="apply_mask_image/output" />
     <rosparam subst_value="true">
       bg_label: -1
+      approximate_sync: $(arg approximate_sync)
```
iory, line 66: How about adding `queue_size`?
The correct usage of `approxiymate_sync` is when you want to get time synchronized data.

poyotamu000: Thank you very much.
Even if I turned off approimate_sync, the viewer worked when I set queue_size to 30 or more.
However, if the queue_size is around 30, the viewer is slow to start up, or only one screen comes up 
(Normally, four screens are activated ... , maybe too slow?), so it is not stable.
Now, for stability, I set the queue_size to 50.

Anyway, why did it work well when I changed approximate_sync?
When approximate_sync is set to be true, does the queue_size also change somewhere?

iory: It's because you made the slop bigger.
Since slop is set to 1, it behaves as if it synchronizes even if it is off by at most 1 second.

Since this is a sample program, it may not be too bad, but it is better to use `Exact synchronize` for inputs (e.g. same sensor sources) that can be synchronized in time.

PR #2625, opened 2021-09-07
add normal publisher and normal marker array

knorth55:
this is from @aginika 's master.


aginika:
May not work well, I think 🤔

PR #2627, opened 2021-09-17
Improved PCL compatibility.

BaltashovIlia:
This MR allows compiling jsk_recognition with PCL 1.11+. It can still be used with previous PCL versions.


knorth55:
thanks, please wait for CI.

PR #2633, opened 2021-10-05
Replace color.pyx with color.py for better compatibility

wkentaro:
**Why?**

Currently, color.pyx is compiled with Cython and it is compatible only with the Python version that is compiled with.  This raises an error when we use virtual environment (e.g., Anaconda) and run `import jsk_recognition_utils` in it (`color` module is imported in `jsk_recognition_utils.__init__`).

**Error message**

When I `import jsk_recognition_utils` from Python 3.7.X on Ubuntu 20.04 (where default Python3 is Python 3.8.10):

```
/home/wkentaro/projectX/.anaconda3/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: compiletime version 3.8 of module 'jsk_recognition_utils.color' does not match runtime version 3.7
return f(*args, **kwds)
Traceback (most recent call last):
File "/home/wkentaro/projectX/ros/src/morefusion/ros/morefusion_panda_ycb_video/nodes/tile_image.py", line 8, in <module>
import jsk_recognition_utils
File "/opt/ros/noetic/lib/python3/dist-packages/jsk_recognition_utils/init.py", line 2, in <module>
from jsk_recognition_utils import color
File "type.pxd", line 9, in init jsk_recognition_utils.color
ValueError: builtins.type size changed, may indicate binary incompatibility. Expected 880 from C header, got 864 from PyObject
```

**Context**

The reason why it was originally implemented with Cython was because of the 2 nested for-loop in `jsk_recognition_utils.color.label_colormap`. However, I reimplemented the same functionality using Numpy operation (no for-loops), so we probably don't need to use Cython anymore.


wkentaro:
Tested with this snippet:
```python
import time

import jsk_recognition_utils


print(jsk_recognition_utils.color)

t_start = time.time()
for _ in range(1000):
    cmap = jsk_recognition_utils.color.labelcolormap()
    cmap.dtype, cmap.shape
print(time.time() - t_start)

```

Results are:

**Before**
```
% python test_label_colormap_performance.py
<module 'jsk_recognition_utils.color' from '/home/wkentaro/Documents/jsk_recognition/devel/lib/python3/dist-packages/jsk_recognition_utils/color.so'>
6.285200119018555
```

**After**
```
% python test_label_colormap_performance.py
<module 'jsk_recognition_utils.color' from '/home/wkentaro/Documents/jsk_recognition/src/jsk_recognition/jsk_recognition_utils/python/jsk_recognition_utils/color.py'>
0.06862044334411621
```

I also tested with pure Python + for-loop:

```
% python test_label_colormap_performance.py
<module 'jsk_recognition_utils.color' from '/home/wkentaro/Documents/jsk_recognition/src/jsk_recognition/jsk_recognition_utils/python/jsk_recognition_utils/color.py'>
7.9616193771362305
```

The reason why we didn't have a big performance gain from Cython here was probably that it has been calling NumPy function in it (np.bitwise_or). I'm not sure exact reason but it shouldn't have been fully exploiting the Cython feature.

wkentaro:
```
diff --git a/jsk_recognition_utils/python/jsk_recognition_utils/color.pyx b/jsk_recognition_utils/python/jsk_recognition_utils/color.pyx
index 0e6b493f..3393116e 100644
--- a/jsk_recognition_utils/python/jsk_recognition_utils/color.pyx
+++ b/jsk_recognition_utils/python/jsk_recognition_utils/color.pyx
@@ -21,10 +21,10 @@ def labelcolormap(int N=256):
         id = i
         r, g, b = 0, 0, 0
         for j in range(0, 8):
-            r = np.bitwise_or(r, (bitget(id, 0) << 7-j))
-            g = np.bitwise_or(g, (bitget(id, 1) << 7-j))
-            b = np.bitwise_or(b, (bitget(id, 2) << 7-j))
-            id = (id >> 3)
+            r |= bitget(id, 0) << 7-j
+            g |= bitget(id, 1) << 7-j
+            b |= bitget(id, 2) << 7-j
+            id = id >> 3
         cmap[i, 0] = r / 255.
         cmap[i, 1] = g / 255.
         cmap[i, 2] = b / 255.
```

With the above change, there's a huge performance improvement:

```
<module 'jsk_recognition_utils.color' from '/home/wkentaro/Documents/jsk_recognition/devel/lib/python3/dist-packages/jsk_recognition_utils/color.so'>
0.7372055053710938
```

PR #2641, opened 2021-12-08
[WIP] add PeoplePose2Dto3D node

knorth55:
@tohirose 

this node convert 2d people pose array to 3d people pose array with depth image and camera info.


nakane11:
I added HumanSkeletonArray for coral as [people_pose_estimation_2d.py](https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_perception/node_scripts/people_pose_estimation_2d.py). I  would like to leave this somewhere because lime names and indices are different from those of OpenPose. I had to make limb_sequence.

However, I wonder if people_pose_2d_to_3d should be common, so should I make new node which coverts 3d pose to skeleton?

knorth55:
if we have prior knowledge, we can convert from pose message to human skelton message.
(i.e. `lhand -> torso is connected`)

nakane11:
As far as I can see [this](https://github.com/jsk-ros-pkg/jsk_recognition/blob/4aee19de0fbc8b3a0619f3043b1a59a1e84e3d3f/jsk_perception/node_scripts/people_pose_estimation_2d.py#L317-L331), bone is added to skeleton if limb_sequence pair exists in vertices.
So we need limb_sequence and index2limbname information beforehand.

I think of coral specific node.

nakane11:
Now I publish skeleton in [people_pose_2d_to_3d](https://github.com/nakane11/teach_spot/blob/d4a513a3baf1aef097d204ee93852dac87cfa08d/node_scripts/people_pose_2d_to_3d.py#L26-L48) in my package.

knorth55:
if we can give this info by rosparam, it would be not specific to coral node.
https://github.com/nakane11/teach_spot/blob/d4a513a3baf1aef097d204ee93852dac87cfa08d/node_scripts/people_pose_2d_to_3d.py#L26-L48

knorth55:
I add doc. this PR is ready for review.

PR #2646, opened 2021-12-21
[jsk_recognition_utils] Enable publishing coefficients array

iory:
# What is this?

Added option to publish coefficients array to `poygon_array_publisher` node. If the polygon is a plane, the coeffiecients array is output in the same way as the output of plane detection.

### DIFF
--- a/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
+++ b/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
@@ -47,6 +47,7 @@ class PolygonArrayPublisher(object):
         self.pub_polygons = rospy.Publisher("~output", PolygonArray, queue_size=1)
         if self.publish_coeffs:
             if self.are_polygons_in_planar is False:
+                self.publish_coeffs = False
                 rospy.logwarn(
                     '~publish_coeffs is True,'
                     ' but ModelCoefficientsArray is not published'
--- a/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
+++ b/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
@@ -11,12 +11,10 @@ from pcl_msgs.msg import ModelCoefficients
 
 def is_planar(points):
     points = np.array(points)
-    # 3 points are always in a plane.
-    if len(points) < 4:
-        return True
     # 2d points are always in a plane.
     if len(points[0]) < 3:
-        return True
+        coeffs = np.array([0, 0, 1, 0], 'f')
+        return True, coeffs
     # Pick out three points (p0, p1, p2) from the given points.
     # Make sure that the three points do not line up in a straight line.
     p0 = points[0]
--- a/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
+++ b/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
@@ -28,6 +28,9 @@ def is_planar(points):
     # Calculate the normal to the plane composed of p0, p1, and p2.
     normal = np.cross(v0, v1)
     norm = np.linalg.norm(normal, ord=2)
+    # If the input polygon is a line or a point, it is not considered a plane.
+    if np.isclose(norm, 0.0):
+        return False, np.array([0, 0, 0, 0], 'f')
     normal = normal / norm
     # Check the normal vector and the vector from p0 to each point are vertical
     dists = np.abs(np.dot(points - p0, normal))
--- a/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
+++ b/jsk_recognition_utils/node_scripts/polygon_array_publisher.py
@@ -17,6 +17,8 @@ def is_planar(points):
     # 2d points are always in a plane.
     if len(points[0]) < 3:
         return True
+    # Pick out three points (p0, p1, p2) from the given points.
+    # Make sure that the three points do not line up in a straight line.
     p0 = points[0]
     points = points[1:]
     dists = np.linalg.norm(points - p0, axis=1, ord=2)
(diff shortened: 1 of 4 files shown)


708yamaguchi (commented):
I have a question about coef value.

When I use sample_polygon_array_publisher.launch, I got the following output.
```
$ roslaunch jsk_recognition_utils sample_polygon_array_publisher.launch
```
Screenshot 2021-12-21 16:41:51
These places can be represented as the following:
- 0\*x + 0\*y + 1*z + 0 = 0
- 0\*x + 0\*y + 1*z + -0.5 = 0
- 0\*x + 0\*y + 1*z + -1 = 0
- 0\*x + -1\*y + 0*z + 1 = 0

So, coefficients [a, b c, d] should be 
- [0, 0, 1, 0]
- [0, 0, 1, -0.5]
- [0, 0, 1, -1]
- [0, -1, 0, 1]

In my understanding, [a, b, c] vector is the same as plane direction (arrow in RViz)


However, I got the following `/polygon_array_publisher/output/coefficients` value.
Is this correct?
```
 rostopic echo /polygon_array_publisher/output/coefficients                                         
header:                                                                                              
  seq: 1                                                                                             
  stamp:                                                                                             
    secs: 1640072311                                                                                 
    nsecs: 461468935                                                                                 
  frame_id: "base_link"                                                                              
coefficients:                                                                                        
  -                                                                                                  
    header:                                                                                          
      seq: 0                                                                                         
      stamp:                                                                                         
        secs: 1640072311                                                                             
        nsecs: 461468935                                                                             
      frame_id: "base_link"                                                                          
    values: [0.0, 0.0, -1.0, -0.0]                                                                   
  -                                                                                                  
    header:                                                                                          
      seq: 0                                                                                         
      stamp:                                                                                         
        secs: 1640072311                                                                             
        nsecs: 461468935                                                                             
      frame_id: "base_link"                                                                          
    values: [0.0, 0.0, -1.0, 0.5]                                                                    
  -                                                                                                  
    header:                                                                                          
      seq: 0                                                                                         
      stamp:                                                                                         
        secs: 1640072311                                                                             
        nsecs: 461468935                                                                             
      frame_id: "base_link"                                                                          
    values: [0.0, 0.0, -1.0, 1.0]                                                                    
  -                                                                                                  
    header:                                                                                          
      seq: 0                                                                                         
      stamp:                                                                                         
        secs: 1640072311                                                                             
        nsecs: 461468935                                                                             
      frame_id: "base_link"                                                                          
    values: [0.0, 1.0, 0.0, -1.0]                                                                    
---                                        
```

jsk_recognition_utils/node_scripts/polygon_array_publisher.py
```diff
@@ -1,21 +1,60 @@
...
+import numpy as np
 import rospy
-from geometry_msgs.msg import PolygonStamped, Point32
+from geometry_msgs.msg import Point32, PolygonStamped
+from jsk_recognition_msgs.msg import ModelCoefficientsArray
 from jsk_recognition_msgs.msg import PolygonArray
+from pcl_msgs.msg import ModelCoefficients
+
+
+def is_planar(points):
```
708yamaguchi, line 12: It would be helpful if you could write a summary or comment about the algorithm.

For example, I can easily understand this web page.
http://www.not-enough.org/abe/manual/argo/pointOnPlane.html

iory: I see. Why don't you write a comment for understanding?

iory:
> In my understanding, [a, b, c] vector is the same as plane direction (arrow in RViz). (I am not confident because I cannot find any documentation about plane direction)

Directions of arrows depend on visualization program. (not using coefficients)
jsk_rviz_plugin's visualization program internally calls jsk_recognition_utils's Plane class.

https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_recognition_utils/src/geo/polygon.cpp#L87-L89
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_recognition_utils/src/geo/plane.cpp#L66-L70

```
 Plane((vertices[1] - vertices[0]).cross(vertices[2] - vertices[0]).normalized(), vertices[0]),
```

The direction of the normal is defined by each program, so it cannot be determined in general.

jsk_recognition_utils/node_scripts/polygon_array_publisher.py
708yamaguchi: OK, I will try.

708yamaguchi:
> The direction of the normal is defined by each program, so it cannot be determined in general.

Thank you for your kind explanation.
I understand.

jsk_recognition_utils/node_scripts/polygon_array_publisher.py
708yamaguchi: I created PR
https://github.com/iory/jsk_recognition/pull/13

708yamaguchi (approved):

708yamaguchi:
I found that tests depending on `polygon_array_publisher.py` fails.
https://github.com/jsk-ros-pkg/jsk_recognition/runs/4602816415?check_suite_focus=true

This is because 
the number of `is_planar`'s return value is expected two
```
planar, coeffs = is_planar(polygon["points"])
```
but sometimes `is_planar` returns single value.
```
    # 3 points are always in a plane.
    if len(points) < 4:
        return True
    # 2d points are always in a plane.
    if len(points[0]) < 3:
        return True
```

```
$ rostest jsk_perception polygon_array_to_label_image.test                                                                                                                                                  
... logging to /home/leus/.ros/log/rostest-leus-ThinkPad-P1-Gen-3-5841.log                                                                                                                                  
[ROSUNIT] Outputting test results to /home/leus/.ros/test_results/jsk_perception/rostest-test_polygon_array_to_label_image.xml                                                                              
[ERROR] [1640177716.496418]: [[1.5, -1.0, 2.0], [-1.0, -1.0, 4.0], [0.5, 1.0, 3.0]]
[ERROR] [1640177716.497147]: True
Traceback (most recent call last):
  File "/home/leus/ros/melodic/src/jsk-ros-pkg/jsk_recognition/jsk_recognition_utils/node_scripts/polygon_array_publisher.py", line 126, in <module>
    p = PolygonArrayPublisher()
  File "/home/leus/ros/melodic/src/jsk-ros-pkg/jsk_recognition/jsk_recognition_utils/node_scripts/polygon_array_publisher.py", line 53, in __init__
    self.msg, self.coeffs_array_msg = self.parse_params(param, frame_id)
  File "/home/leus/ros/melodic/src/jsk-ros-pkg/jsk_recognition/jsk_recognition_utils/node_scripts/polygon_array_publisher.py", line 99, in parse_params
    planar, coeffs = is_planar(polygon["points"])
TypeError: 'bool' object is not iterable
^C^C^C[Testcase: testtest_polygon_array_to_label_image] ... ok
```

708yamaguchi (changes requested):
Sorry, the previous comment is request change.

iory:
Thanks.
I fixed the returning value.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2646/commits/14ab5ed3c99661332d67b691d7332d683e30d72f

I have also addressed the error when the input polygons are points or lines.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2646/commits/12ad4d4ac99f7783167ad917a7cfa2e0d26a2914

708yamaguchi (approved):

PR #2647, opened 2021-12-21
[jsk_recognition_utils] Add polygon_array_to_box_array node

iory:
# What is this?

Convert `jsk_recognition_msgs/PolygonArray` to `jsk_recognition_msgs/BoundingBoxArray` containing the polygons. 
For example, this can be used when users use polygon for collision avoidance programs.

image


708yamaguchi (commented):
Thank you.
I think this is very good feature.

I can use this PR easily.
- bounding box array nicely enclose polygon
- dynamic reconfigure works well.

If you could write an document or comment of the algorithm in `polygon_array_publisher.py`, it would be helpful for me to understand it.
For example, why do we need `rotate_points` and `rotation_matrix_from_axis`? Is this to use `shapely`?

708yamaguchi (changes requested):
Tests fails similar to https://github.com/jsk-ros-pkg/jsk_recognition/pull/2646#issuecomment-999580534

iory:
> If you could write an document or comment of the algorithm in polygon_array_publisher.py, it would be helpful for me to understand it.

I added comments.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2647/commits/3dd875c35c0c649cbf8e898fb02f5549639524f0

708yamaguchi:
Indigo and Kinetic tests fail because `python-numpy` rosdep key is removed.

https://github.com/ros/rosdistro/pull/31552

PR #2649, opened 2021-12-23
Add python-numpy due to ros/rosdistro#31580

tkmtnt7000:
This is in response to https://github.com/ros/rosdistro removing all EOL keys related to python-numpy in https://github.com/ros/rosdistro/pull/31552.
I used https://github.com/jsk-ros-pkg/jsk_3rdparty/pull/306/commits/fcdd7ce9b720124e97c419e388a4f81ae85f13ec as a reference for writing.

### DIFF
--- a/.github/workflows/indigo.yml
+++ b/.github/workflows/indigo.yml
@@ -24,6 +24,7 @@ jobs:
         uses: jsk-ros-pkg/jsk_travis@master
         with:
           USE_JENKINS: true
+          EXTRA_DEB : "python-numpy"
           ROS_DISTRO : indigo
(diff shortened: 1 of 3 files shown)


k-okada:
@tkmtnt7000 If https://github.com/ros/rosdistro/pull/31552 is their final decision, we will update jsk_travis to use old rosdep key for indigo/kinetic. so...
Ideal situation, they change mind and keep old rosdep key.
Best solution, fix jsk_travis and enable all jsk-ros-pkg to work with deleted rosdep key
Clever solution (specially, if you have deadline), wait

tkmtnt7000:
Thank you for advice.
I'll wait.

PR #2656, opened 2022-01-06
option to use ClusterPointIndices for outlier removal

knorth55:
Remove outlier with cluster point indices.
With this PR, we can run outlier removal in parallel with ClusterPointIndices message.

- [x] doc
- [x] sample

BBox without outlier removal: Edge bbox
BBox with outlier removal: Color bbox
Screenshot from 2022-01-07 03-06-25


knorth55:
I also update the doc.

knorth55:
I update this PR to support not only `XYZRGB` but also other formats such as `XYZI`.

doc/jsk_pcl_ros/nodes/organized_statistical_outlier_removal.md
```diff
@@ -1,4 +1,5 @@
-# OrganizedStatisticalOutlierRemoval 
+# OrganizedStatisticalOutlierRemoval
```
iory, line 1: Overall, this PR seems good. One thing I would like to add is the distinction between the usage of Euclidean clustering and this node. Both can be given the same input, and while Euclidean clustering outputs cluster indices, it can ultimately produce a point cloud, making them quite similar in practice. It would be helpful if the documentation could include guidance on how users should differentiate between using this node and Euclidean clustering.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/3ed4aa6c7fdab03c97d208b7021b2f8f2a1104f1/doc/jsk_pcl_ros/nodes/euclidean_clustering.md

iory (changes requested):

doc/jsk_pcl_ros/nodes/organized_statistical_outlier_removal.md
knorth55: I intended to use this after Euclidean clustering.
Euclidean clustering makes the clusters with euclidean distance,
and this node can remove the shadow of the points from clusters.
Removing shadow by just using Euclidean clustering is a bit hard, because it does not consider the distribution.

iory: OK. Can you write the document in a way that many users can understand? Additionally, I would like you to include an explanation with a link to Euclidean clustering.

knorth55: I added sample launch for this feature, and i think it is enough.
I don't get what you want me to write about it.
Can you clearly tell me what you want me to write?

iory: I believe that many of the people who will use this are not developers but rather users.
Among them, few will be able to understand just by looking at the usage of the launch file.

I think many users (including developers) will appreciate it if you could write documentation that even fourth-year undergraduate students can understand.
Could you add and revise the documentation as you answered my questions?

knorth55: Added.

knorth55:
Rebased, and solved conflicts.
Now docs are updated, so ready to be reviewed.

PR #2659, opened 2022-01-08
chmod -x sound_classification scripts for catkin_virtualenv

knorth55:
because `scripts/*.py` is installed by `catkin_virtualenv`, it should be `644` not `755` to determine the executable.


708yamaguchi:
Thank you very much for your PR.

I think this change is recommended.

Even if we set the scripts to 755, `catkin_virtualenv` will change it to 644 in `catkin build`.
https://github.com/locusrobotics/catkin_virtualenv/blob/71342edb9d39339dfb639f1022704cff954db60d/catkin_virtualenv/cmake/catkin_install_python.cmake#L51-L57

PR #2663, opened 2022-01-19
use default value for param

knorth55:
use default value for `rosparam` to prevent load error.


iory (approved):

soonhyo:
Attention clipper sometimes fails with this error.
I'm testing this PR and checking if this PR solve the following error.

```
[FATAL] [1651121652.676082280]: ASSERTION FAILED
	file = /opt/ros/melodic/include/message_filters/sync_policies/approximate_time.h
	line = 122
	cond = queue_size_ > 0
```

PR #2666, opened 2022-02-03
[jsk_pcl_ros] Add crop points from bounding box array for Attention clipper sample

iory:
# What is this?

Added a sample to crop a point cloud from the input of a bounding box array using the attention clipper node.

## quickstart

Build `jsk_pcl_ros` and run following command.
```
roslaunch jsk_pcl_ros sample_attention_clipper.launch
```


k-okada:
Could you update doc? https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/doc/jsk_pcl_ros/nodes/attention_clipper.md

PR #2667, opened 2022-02-21
Use DiagnosticTransport instead of ConnectionBasedTransport

iory:
# What is this?

This PR depends on this jsk_common's [PR](https://github.com/jsk-ros-pkg/jsk_common/pull/1711).
With this change, users can see if the topics are properly connected and processing is done by diagnostics topic.


iory:
This PR is no longer needed as the diagnostics feature has been incorporated into the connectionbased transport.

PR #2668, opened 2022-03-07
[WIP] [jsk_perception] Add removing background node using rembg

iory:
# What is this?

This PR added a node to remove the background using [rembg](https://github.com/danielgatis/rembg).

## Some problems 

Users using versions below melodic need to install python3.8 to meet dependencies such as numpy.
You also need to build cv_bridge from source so that it can be run on python3.

# Quick start

```
catkin build jsk_perception
```
Source after this and run the following command.

```
roslaunch jsk_perception sample_remove_background.launch
```


k-okada:
@iory can you check we can use python3 version of `cv_bridge` on melodic with https://drive.google.com/file/d/1KOSdtFt0WEYC-qLc66-18L0XH_Av2f2a/view?usp=sharing ??
you need to add 
```
import sys, os; sys.path.insert(0,'/opt/ros/' +  os.environ['ROS_DISTRO'] + '/lib/python3/dist-packages/')
```
before you call
```
import cv_bridge
```

If this fails, please provide me how to reproduce this.

iory:
Thank you for preparing cv_bridge for python3.
I have confirmed that cv_bridge works well in melodic's python3 environment.

k-okada:
Memo : with @iory 
```
jsk_perception_3rdparty/
                          3rdparty/rembg (submodule)
                          node_script/rembg_node.py 
                          CMakeLists.txt (only support melodic+ )
                          package.xml
```

PR #2669, opened 2022-03-16
[jsk_perception] Add libuvc_camera to dependency for sample_insta360_air.launch

tkmtnt7000:
`sample_insta360_air.launch` use `libuvc_camera` but no dependency is written in `package.xml`.
So, we have to execute `sudo apt install ros-melodic-libuvc-camera` manually without this pull request.

Please let me know if there is any reason why `libuvc_camera` is not listed in the dependencies.

cc:@708yamaguchi

### DIFF
--- a/jsk_perception/package.xml
+++ b/jsk_perception/package.xml
@@ -32,6 +32,7 @@
   <build_depend>jsk_recognition_utils</build_depend>
   <build_depend>jsk_topic_tools</build_depend>
   <build_depend>libcmt</build_depend>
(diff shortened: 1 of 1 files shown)


knorth55 (approved):

708yamaguchi:
@tkmtnt7000 

Thank you very much for your PR.

When releasing jsk_perception, I thought it would be better to have fewer dependent packages.

If there are no release issues, I think this pull request is necessary.

PR #2672, opened 2022-04-07
Enable 3d hand pose detection

nakane11:
Updated hand pose estimation to get hand pose in 3d．
Since the previous rosbag data doesn't include depth image, I replaced it with new one compatible．

`roslaunch jsk_perception sample_hand_pose_estimation_2d.launch GPU:=-1 with_depth:=true`

Cc: @iory

### DIFF
--- a/doc/jsk_perception/nodes/hand_pose_estimation_2d.rst
+++ b/doc/jsk_perception/nodes/hand_pose_estimation_2d.rst
@@ -1,4 +1,4 @@
-hand_pose_estimation_2d.py
+oWANG-SRH-2019-07.hand_pose_estimation_2d.py
 ==========================
 
 
@@ -20,19 +20,33 @@ Subscribing Topic
 
   Input image.
 
+* ``~input/depth`` (``sensor_msgs/Image``)
 
+  Input depth image.
+
+* ``~input/info`` (``sensor_msgs/CameraInfo``)
+
+  Input camera info.
+
+  
 Publishing Topic
 ----------------
-
-* ``~output/pose`` (``jsk_recognition_msgs/HandPoseArray``)
-
-  Detected hand keypoints 2D positions and scores in image.
 
 * ``~output/vis`` (``sensor_msgs/Image``)
 
   Visualization image of detected hand poses.
 
+* ``~output/pose`` (``jsk_recognition_msgs/HandPoseArray``)
 
+  If ``with_depth`` is true, publish 3D joint position.
+
+  If ``with_depth`` is false, publish 2D joint position in image.
+
+* ``~output/pose2d`` (``jsk_recognition_msgs/HandPoseArray``)
+
+  If ``with_depth`` is true, publish 2D joint position.
+
+  
 Parameters
 ----------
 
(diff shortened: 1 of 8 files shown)


tkmtnt7000:
Nice PR!
I think it is very good feature. 

Maybe I am overthinking this, but I'm a little concerned about the potential for confusion with file names.
The node file name is `hand_pose_estimation_"2d".py`, but this node gives us "3d" hand pose in addition to 2d hand pose.

Affonso-Gui:
(was thinking the same thing, but am not sure which would be the right way to proceed. Maybe keeping the `hand_pose_estimation_2d.py` and adding a new `hand_pose_estimation_3d.launch` and `sample_hand_pose_estimation_3d.launch`? Or just removing the "2d" part from the file name altogether if it doesn't cause any compatibility issues..)

iory:
This node originally estimates the pixel position of hands. 
In my opinion, estimating the 3d position is an optional feature of this node.

> Maybe keeping the hand_pose_estimation_2d.py and adding a new hand_pose_estimation_3d.launch and sample_hand_pose_estimation_3d.launch?

So we don't have to erase `2d` and I agree with this idea.

knorth55:
Good PR!
IMO, there is another solution for further development.
I think it might be useful to separate this feature to one node like `HandPose2Dto3D`,
which subscribe 2D `HandPoseArray`, `DepthImage`, `CameraInfo` and publish 3d `HandPoseArray`.
But, this PR looks good to me.

FYI:
I made `PeoplePose2Dto3D` to separate the simliar feature.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2641

Affonso-Gui:
@knorth55 Also seems nice, specially considering we have the `PeoplePose2Dto3D` precedent

doc/jsk_perception/nodes/hand_pose_estimation_2d.rst
```diff
@@ -20,19 +20,33 @@ Subscribing Topic
...
 * ``~output/pose`` (``jsk_recognition_msgs/HandPoseArray``)
 
-  Detected hand keypoints 2D positions and scores in image.
+  If ``with_depth`` is true, publish 3D joint position.
 
-* ``~output/vis`` (``sensor_msgs/Image``)
+  If ``with_depth`` is false, publish 2D joint position in image.
 
-  Visualization image of detected hand poses.
+* ``~output/pose2d`` (``jsk_recognition_msgs/HandPoseArray``)
```
iory, line 45: ``~output/pose_2d``  is correct.

iory (changes requested):

doc/jsk_perception/nodes/hand_pose_estimation_2d.rst
nakane11: I revised it.

iory (approved):

doc/jsk_perception/nodes/hand_pose_estimation_2d.rst
iory: Nice!

PR #2676, opened 2022-04-19
[jsk_perception] return when camera info message is None

knorth55:
this PR avoids error when `self.cameramodel` is not set.


iory (approved):

iory:
Test passed. I think this patch is small and good to merge.

knorth55:
this PR passed all required tests.

PR #2677, opened 2022-04-21
work around permission issue / integrate all yaml to one

k-okada:
as github finally support re-run failed jobs, https://github.blog/2022-03-16-save-time-partial-re-runs-github-actions/#:~:text=If%20you%20have%20failing%20jobs,directly%20from%20the%20logs%20view.


k-okada:
noetic test fails as expected

https://gist.github.com/k-okada/6a9cd394e7e82241745cd7b2049e9122

PR #2679, opened 2022-04-26
publish boundingboxarray in paper_finder.py

YUKINA-3252:
Before this change, paper_finder.py did not publish the size of paper.
This PR enables this to publish the information of the size as BoundingBoxArray.

cc. @iory @knorth55


iory (commented):
Good PR!

Let's have the bounding box array displayed in rviz launched by `sample_paper_finder.launch` so that others can easily see the result of this change.

Also, let's add a bounding box array to the document below.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/doc/jsk_perception/nodes/paper_finder.md#publishing-topics

iory (approved):

PR #2681, opened 2022-05-04
Add reset sync policy in destructor and add a test

iory:
# What is this?

Some nodelets using message filters raise the following error when the program is killed.
This error kills the nodelet manager itself, killing all plugins registered in the nodelet.

```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
Aborted
```
This PR fixes these errors and enables tests on whether to publish the topic again after killing the rosnode.

Related to https://github.com/jsk-ros-pkg/jsk_common/pull/1640 https://github.com/ros/ros_comm/issues/720 

cc: @nakane11 @708yamaguchi


k-okada (changes requested):
please add jsk_recognition_utils/test/test_topic_published.py to jsk_topics, if this is simply updated verson of jsk_tools/src/test_topic_published.py

jsk_pcl_ros_utils/src/colorize_distance_from_plane_nodelet.cpp
```diff
@@ -57,6 +57,10 @@ namespace jsk_pcl_ros_utils
     srv_->setCallback (f);
   }
 
+  ColorizeDistanceFromPlane::~ColorizeDistanceFromPlane() {
```
k-okada, line 60: If  we basically add `sync_.reset()` to all, how about define someting like
```
  class Recognition?Filter?Nodelet???: public jsk_topic_tools::DiagnosticNodelet {
       boost::shared_ptr<message_filters::Synchronizer<SyncPolicy> > sync_;
       Recognition?Filter?Nodelet???() {
               sync_ = boost::make_shared<message_filters::Synchronizer<SyncPolicy> >(100);
      }
      ~Recognition?Filter?Nodelet???() {
           sync_.reset(); 
      }
  };

  class PolygonPointsSampler: public jsk_pcl_ros_utils::Recognition?Filter?Nodelet???
```
?? may be you already considered this, and find some drawbacks.

jsk_pcl_ros/src/extract_indices_nodelet.cpp
```diff
@@ -58,6 +58,19 @@ namespace jsk_pcl_ros
     onInitPostProcess();
   }
 
+  ExtractIndices::~ExtractIndices() {
+    // This fixes the following error on shutdown of the nodelet:
+    // terminate called after throwing an instance of
+    // 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
+    //     what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
+    // Also see https://github.com/ros/ros_comm/issues/720 .
```
k-okada, line 66: Overall, this PR looks nice, my conceren is that, if some will create new nodelet in future and they did not know about this problem&fixes, and run into same problem, so I think we should add this comment to all nodes, or impliment this fix to all nodelet implcitly. 
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2681#discussion_r865521568

jsk_pcl_ros_utils/src/colorize_distance_from_plane_nodelet.cpp
iory: I have considered it. 
However, some programs can have multiple asyncs and syncs.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/659a25c7bed3c8786aefa6ef1ce16a9bd53aaecb/jsk_pcl_ros/include/jsk_pcl_ros/ppf_registration.h#L91-L93
We believe that defining a class like this reduces flexibility.
So I added a comment to all programs that use synchronizer.

iory:
> please add jsk_recognition_utils/test/test_topic_published.py to jsk_topics

Thank you for your comments.
I send a PR to add a modified `test_topic_published.py`.
https://github.com/jsk-ros-pkg/jsk_common/pull/1725

knorth55 (approved):

iory:
I changed to use this test https://github.com/jsk-ros-pkg/jsk_common/pull/1725 .
@k-okada 
This pr focuses on bug fixes by adding tests that check topics after killing nodes.
Test passed.
If there is no problem, please merge this PR.

Issue #2685, opened 2022-05-13
Regression caused by update objectdetection_tf_publisher for publishing simple tf #457


amirtu:
Could someone place the line https://github.com/jsk-ros-pkg/jsk_recognition/blob/39747cc75031e3ff599450189a3f3645b31145a5/checkerboard_detector/src/objectdetection_tf_publisher.py#L19
after these two lines
https://github.com/jsk-ros-pkg/jsk_recognition/blob/39747cc75031e3ff599450189a3f3645b31145a5/checkerboard_detector/src/objectdetection_tf_publisher.py#L20-L21
as it was before [#457](https://github.com/jsk-ros-pkg/jsk_recognition/commit/c5478102d20e9b946f47277347698e9c1c64ddb8)

knorth55:
You can make a PR.
This repo is open-source, so you can join by yourself.
Don't hesitate and please feel free to submit your suggestion and describe the reason why you want it in a PR.

amirtu:
@knorth55 

Thanks for your prompt reply, I have created a PR for this change.

knorth55:
can you explain the detailed error you encountered?
can you also paste the error output?

PR #2686, opened 2022-05-13
Fix regression

amirtu:
As to resolve the #2685


k-okada:
@amirtu Thank you for contribution. Can you explain the original error? (and if you provide test code, that welcome!)

PR #2688, opened 2022-05-17
Use boost::placeholders::_N instead of deprecated _N, also missing dependency on diagnostic_updaters

lucasw:
These changes made this build in Ubuntu 22.04 (except for sound_classification, I think there is a catkin virtualenv issue with that) and reduced the amount of build warnings (there are also a few `warning: no return statement in function returning non-void` to clean up later)


lucasw:
I just rebased this on the latest `master`

lucasw:
Made a clean new commit here on top of the latest master to get rid of the merge history

PR #2690, opened 2022-05-27
[jsk_perception/dualfisheye_panorama] poke in callback

iory:
# What is this?

In the current implementation, `DualFisheyeToPanorama` node does not poke in callback loop.
This is a simple change to make poke.

cc: @708yamaguchi @tkmtnt7000


708yamaguchi:
Sorry I already made PR like this.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2684

Move to this PR.

PR #2697, opened 2022-06-11
 [jsk_perception/aws_detect_faces] Publish visualized image when use_image is False case

iory:
# What is this?

In current, If `use_window` is not set to true, an error will occur because the visualized image has not been initialized.
This PR fix the bug.

Also, this pr initialize `img_width` and `img_height` at callback loop.
Even when `use_window` is false, `img_gray.shape` containing `None` may be used. I also fixed this bug.


jsk_perception/node_scripts/aws_detect_faces.py
```diff
@@ -341,11 +349,12 @@ def image_callback(self, image):
             if pose_msgs:
                 pose_msgs.poses.append(pose_msg)
 
-        if self.use_window:
+        if visualize:
```
k-okada, line 352: This should be `self.use_window`? because we do not need to call `imshow` and when we just publish images,

iory: Thank you for your comments.
I modified it to separate cv2's imshow and image publishing.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2697/commits/0813d5ee7d4f583054d025e2e83c633fe1642e00

Issue #2706, opened 2022-06-30
How to Invert a Mask Image?


Affonso-Gui:
How to invert a mask image in the jsk_recognition pipeline?
I have the following:
![mask](https://user-images.githubusercontent.com/20625381/176597169-a8c14666-573f-4d42-afa0-fd3a83c71afc.png)

And want this:
![mask_invert](https://user-images.githubusercontent.com/20625381/176597235-cd0cb1aa-79d9-4f04-8d59-419e5cf63e37.png)

iory:
You can use `ApplyMaskImage` with `~negative` option.
https://jsk-docs.readthedocs.io/projects/jsk_recognition/en/latest/jsk_perception/nodes/apply_mask_image.html

Affonso-Gui:
Thanks, it worked for me with `negative: true` and `clip: false`, although with significant drops in performance (10Hz -> 6Hz).

Maybe we could also add a negative option directly to the RobotToMask node?

iory:
> Maybe we could also add a negative option directly to the RobotToMask node?

I think the general direction is to create a node that inverts the mask.

Affonso-Gui:
Looking forward to an InvertMask node.

PR #2707, opened 2022-06-30
[jsk_perception/sample/apply_mask] Add clip and negative option for users

iory:
Related to https://github.com/jsk-ros-pkg/jsk_recognition/issues/2706
@Affonso-Gui Could you try and review this PR?


Affonso-Gui:
Do you think it is important to add the options to the sample program? I am quite happy with minimal sample files.

iory:
I think it is useful for users to change with argument when they want to try such a function immediately.

PR #2710, opened 2022-07-06
[jsk_pcl_ros/organized_multi_plane_segmentation_nodelet] Poke when start subscribing

iory:
# What is this?

This commit is related to https://github.com/jsk-ros-pkg/jsk_common/pull/1735
If the `organized_multi_plane_segmentation` is currently subscribed, but the publish function is not executed (`vital_checker->poke()` is not executed), it publishes diagnostics with an error level.
This publishes error level diagnostics only at the beginning because poke is not called when subscribing starts.

### DIFF
--- a/jsk_pcl_ros/include/jsk_pcl_ros/organized_multi_plane_segmentation.h
+++ b/jsk_pcl_ros/include/jsk_pcl_ros/organized_multi_plane_segmentation.h
@@ -187,6 +187,16 @@ namespace jsk_pcl_ros
     bool estimate_normal_;
     bool publish_normal_;
 
+    /** @brief
+     * Previous checked status of connection.
+     */
+    jsk_topic_tools::ConnectionStatus previous_checked_connection_status_for_normal_;
+
+    /** @brief
+     * Previous checked status of connection for plane segmentation.
+     */
+    jsk_topic_tools::ConnectionStatus previous_checked_connection_status_for_plane_;
+
     ////////////////////////////////////////////////////////
     // parameters for RANSAC refinement
     ////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


iory:
At first glance, diff seems to be a lot, but most of them are indents for if-then.
You can see a simple diff with `w=1` option.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2710/files?w=1

708yamaguchi:
Thank you for your improvement!

Why this pull request is needed while https://github.com/jsk-ros-pkg/jsk_common/pull/1735 has already been merged is that 
- `organized_multi_plane_segmentation` does not use `vital_checker_` variable defined in https://github.com/jsk-ros-pkg/jsk_common/blob/714713378fb037c9670bf48118010f1899c10bd5/jsk_topic_tools/include/jsk_topic_tools/diagnostic_nodelet.h#L111
- Instead, `organized_multi_plane_segmentation` uses the original two vital checkers https://github.com/jsk-ros-pkg/jsk_recognition/blob/9bdbc84ca76d1dbd0fecd88f40b6cb47438ab9df/jsk_pcl_ros/include/jsk_pcl_ros/organized_multi_plane_segmentation.h#L169-L170

P.S.
Thanks for sharing such a useful option.
> You can see a simple diff with `w=1` option.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2710/files?w=1

708yamaguchi:
We confirmed the problem by running the following code within the pr2 PC.

https://gist.github.com/708yamaguchi/e95aeb65256de0fb6d740fd6b9f26a69

708yamaguchi (approved):

PR #2711, opened 2022-07-09
[jsk_perception/virtual_camera_mono] Use DiagnosticNodelet to start subscribing topics when camera_info is subscribed

nakane11:
Currently this node starts subscribing input topics if output image is subscribed.
~~I added `info_connect_cb` and `info_disconnect_cb` to set them to advertiseCamera defined in image_transport.~~ 

This node is useful to project robot model from overhead onto the ground. I only subscribe virtual_camera_info.
https://github.com/nakane11/teach_spot/blob/f7bdb84058f14dff8dbbe1799ff6e05dbb447b1c/launch/dynamic_footprint.launch#L11-L22

Please refer to https://github.com/jsk-ros-pkg/jsk_common/issues/1198
Another solution is to use [advertiseCamera](https://github.com/jsk-ros-pkg/jsk_common/blob/30cdff0624b253d58af2b00ccdf0277cf27d9ec1/jsk_topic_tools/include/jsk_topic_tools/connection_based_nodelet.h#L314-L341) in ConnectionBasedNodelet (DiagnosticNodelet inherits it), ~~but I don't think we should do that for backward compatibility.~~ 
Deprecated advertiseCamera passes `info_connect_cb` and `info_disconnect_cb` to original advertiseCamera of image_transport.
cc: @iory

### DIFF
--- a/jsk_perception/src/virtual_camera_mono.cpp
+++ b/jsk_perception/src/virtual_camera_mono.cpp
@@ -1,70 +1,35 @@
-#include <ros/ros.h>
-#include <dynamic_reconfigure/server.h>
 #include <jsk_topic_tools/log_utils.h>
 #include <jsk_topic_tools/rosparam_utils.h>
-#include <image_transport/image_transport.h>
-#include <image_geometry/pinhole_camera_model.h>
-#include <tf/transform_listener.h>
-#include <tf/transform_broadcaster.h>
 #include <cv_bridge/cv_bridge.h>
-#if ( CV_MAJOR_VERSION >= 4)
-#else
-#include <opencv/cv.hpp>
-#include <opencv/highgui.h>
-#endif
-
 #include <string>
 #include <vector>
-#include <boost/assign.hpp>
-#include <boost/thread.hpp>
-
 #include <geometry_msgs/Polygon.h>
-#include <geometry_msgs/PolygonStamped.h>
 
-#include <jsk_perception/VirtualCameraMonoConfig.h>
-
-
-class VirtualCameraMono
+#include "jsk_perception/virtual_camera_mono.h"
+namespace jsk_perception
 {
-  ros::NodeHandle nh_,private_nh_;
-  boost::shared_ptr<dynamic_reconfigure::Server<jsk_perception::VirtualCameraMonoConfig> > srv_;
-  image_transport::ImageTransport it_,it_priv_;
-  image_transport::CameraSubscriber sub_;
-  image_transport::CameraPublisher pub_;
-  int subscriber_count_;
-  tf::TransformListener tf_listener_;
-  image_geometry::PinholeCameraModel cam_model_;
-  tf::TransformBroadcaster tf_broadcaster_;
-  ros::Subscriber sub_trans_, sub_poly_;
-
-  tf::StampedTransform trans_; // transform to virtual camera
-  geometry_msgs::PolygonStamped poly_; // target polygon to transform image
-  int interpolation_method_;
-
-public:
-  VirtualCameraMono() : private_nh_("~"), it_(nh_), it_priv_(private_nh_), subscriber_count_(0)
+  void VirtualCameraMono::onInit()
   {
-    image_transport::SubscriberStatusCallback connect_cb = boost::bind(&VirtualCameraMono::connectCb, this, _1);
-    image_transport::SubscriberStatusCallback disconnect_cb = boost::bind(&VirtualCameraMono::disconnectCb, this, _1);
-    pub_ = it_priv_.advertiseCamera("image", 1, connect_cb, disconnect_cb);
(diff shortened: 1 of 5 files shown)


iory:
> Another solution is to use [advertiseCamera](https://github.com/jsk-ros-pkg/jsk_common/blob/30cdff0624b253d58af2b00ccdf0277cf27d9ec1/jsk_topic_tools/include/jsk_topic_tools/connection_based_nodelet.h#L314-L341) in ConnectionBasedNodelet, but I don't think we should do that for backward compatibility.

Could you explain it to me? I feel that it is a good idea to use the connection based function well.

nakane11:
I don't know much, but if ConnectionBasedNodelet is extended, launch files running this node have to be changed.

iory:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/09cbce5f0ce7102331f6e48b9f9befc2b1fda869/jsk_perception/CMakeLists.txt#L217-L223
https://github.com/jsk-ros-pkg/jsk_common/blob/30cdff0624b253d58af2b00ccdf0277cf27d9ec1/jsk_topic_tools/cmake/nodelet.cmake

In CMakeLists, load and standalone can be written.
We can maintain backward compatibility.

nakane11:
Thank you. I fixed to extend connection based nodelet.
I checked sample_virtual_camera_mono.launch.

jsk_perception/src/virtual_camera_mono.cpp
```diff
@@ -1,70 +1,34 @@
...
-
-public:
-  VirtualCameraMono() : private_nh_("~"), it_(nh_), it_priv_(private_nh_), subscriber_count_(0)
+  void VirtualCameraMono::onInit()
   {
-    image_transport::SubscriberStatusCallback connect_cb = boost::bind(&VirtualCameraMono::connectCb, this, _1);
-    image_transport::SubscriberStatusCallback disconnect_cb = boost::bind(&VirtualCameraMono::disconnectCb, this, _1);
-    pub_ = it_priv_.advertiseCamera("image", 1, connect_cb, disconnect_cb);
+    DiagnosticNodelet::onInit();
+    pub_ = advertiseCamera(*pnh_, "image", 1, false);
```
iory, line 13: `pub_ = advertiseCamera(*pnh_, *it_, "image", 1);` is correct?

iory (changes requested):

jsk_perception/src/virtual_camera_mono.cpp
iory: This method is deprecated. I was misunderstanding.

jsk_perception/src/virtual_camera_mono.cpp
```diff
@@ -111,39 +76,22 @@ class VirtualCameraMono
...
   {
     ROS_INFO("Unsubscibing from image topic");
     sub_.shutdown();
   }
 
-  void imageCb(const sensor_msgs::ImageConstPtr& image_msg,
+  void VirtualCameraMono::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
                const sensor_msgs::CameraInfoConstPtr& info_msg)
   {
     cv_bridge::CvImagePtr cv_ptr;
```
iory, line 97: Could you add `vital_checker_->poke()`?

iory (changes requested):

jsk_perception/src/virtual_camera_mono.cpp
nakane11: Thank you, I fixed it.

jsk_perception/src/virtual_camera_mono.cpp
iory, line 13: https://github.com/jsk-ros-pkg/jsk_recognition/runs/7291655426?check_suite_focus=true
Test failed.
Since jsk_common is released only up to 2.2.11 in kinetic etc., it is better to use `advertiseCamera(ros::NodeHandle& nh, const std::string& topic, int queue_size)`.
The `latch` option is `false` by default, so there is no problem.

iory (changes requested):

tkmtnt7000:
If possible, could you please comment in the title or PR comment, (or in source codes) about what we can do with DiagnosticNodelet and how you came to use DiagnosticNodelet so that we can easily search for it later?

jsk_perception/src/virtual_camera_mono.cpp
nakane11: Thank you! Updated.

nakane11:
@tkmtnt7000 I updated the title and the first PR comment. Is this enough?

tkmtnt7000:
@nakane11 
Thank you for updating! I clearly understand what this PR can do.

iory (approved):

PR #2712, opened 2022-07-11
Use DiagnosticNodelet instead of ConnectionBasedNodelet.

iory:
Modified all `ConnectionBasedNodelets` in jsk_recognition to `DiagnosticNodelet`.

```
$ cd jsk_recognition
$ grep ConnectionBasedNodelet -l -r --include='*.h' . | wc
0       0       0
$ grep ConnectionBasedNodelet -l -r --include='*.hpp' . | wc
0       0       0
```

Also, DiagnosticsNodelet requires poke.
There are many PRs out there that add `poke`, and this PR also added `poke` for all nodelets that do not have `poke`.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2690
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2710
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2203

```
$ cd jsk_recognition
$ grep 'DiagnosticNodelet::onInit' -r . -l | xargs grep -L poke | sort
./.git/logs/HEAD
./.git/logs/refs/heads/reset-sync-policy
./resized_image_transport/src/image_resizer_nodelet.cpp
./resized_image_transport/src/log_polar_nodelet.cpp
```

The nodes of `resized_image_transport` inherit from `image_processing_nodelet`, and `poke` is called inside `image_processing_nodelet`.
https://github.com/jsk-ros-pkg/jsk_recognition/blob/09cbce5f0ce7102331f6e48b9f9befc2b1fda869/resized_image_transport/src/image_processing_nodelet.cpp#L141-L157

### DIFF
--- a/jsk_pcl_ros/include/jsk_pcl_ros/bilateral_filter.h
+++ b/jsk_pcl_ros/include/jsk_pcl_ros/bilateral_filter.h
@@ -38,7 +38,18 @@
 #define JSK_PCL_ROS_BILATERAL_FILTER_H_
 
 #include <pcl_ros/publisher.h>
-#include <jsk_topic_tools/diagnostic_nodelet.h>
+#include <jsk_recognition_utils/jsk_topic_tools_version.h>
+#if JSK_TOPIC_TOOLS_VERSION_MINIMUM(2,2,13)
+  #include <jsk_topic_tools/diagnostic_nodelet.h>
+  namespace jsk_topic_tools {
+    #define NODELET DiagnosticNodelet
+  }
+#else
+  #include <jsk_topic_tools/connection_based_nodelet.h>
+  namespace jsk_topic_tools {
+    #define NODELET ConnectionBasedNodelet
+  }
+#endif
 #include "jsk_recognition_utils/pcl_conversion_util.h"
 #include <jsk_pcl_ros/BilateralFilterConfig.h>
 #include <dynamic_reconfigure/server.h>
@@ -74,13 +85,13 @@ namespace jsk_pcl_ros
   }
 
   
-  class BilateralFilter: public jsk_topic_tools::DiagnosticNodelet
+  class BilateralFilter: public jsk_topic_tools::NODELET
   {
   public:
     typedef boost::shared_ptr<BilateralFilter> Ptr;
     typedef pcl::PointXYZRGB PointT;
     typedef BilateralFilterConfig Config;
-    BilateralFilter() : DiagnosticNodelet("BilateralFilter") {}
+    BilateralFilter(){}
   protected:
     ////////////////////////////////////////////////////////
     // methods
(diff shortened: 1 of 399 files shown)


jsk_pcl_ros/include/jsk_pcl_ros/bilateral_filter.h
```diff
@@ -38,7 +38,7 @@
 #define JSK_PCL_ROS_BILATERAL_FILTER_H_
 
 #include <pcl_ros/publisher.h>
-#include <jsk_topic_tools/connection_based_nodelet.h>
+#include <jsk_topic_tools/diagnostic_nodelet.h>
```
k-okada, line 41: if this code depends on jsk_topic_tools >= x.x.x, I would recommend to write something like https://github.com/PR2/pr2_mechanism/commit/c2d2d7e0f18dad9f82f4f22fbc681d02b3aa6469#diff-b32b543817fc81911592c454ab29c1d29f6ce7cd2ab6dcd3a374014568a2ee55R81-R85

instead of using  `-DJSK_TOPIC_TOOLS_X_X_XAPI` , will we introduce jsk_topic_tools_version.h ???(https://github.com/jsk-ros-pkg/jsk_common/blob/714713378fb037c9670bf48118010f1899c10bd5/jsk_topic_tools/CMakeLists.txt#L30-L32)

jsk_pcl_ros/include/jsk_pcl_ros/bilateral_filter.h
```diff
@@ -74,12 +74,13 @@ namespace jsk_pcl_ros
   }
 
   
-  class BilateralFilter: public jsk_topic_tools::ConnectionBasedNodelet
```
k-okada, line 77: how about
```
#if  defined(JSK_TOPIC_TOOLS_VERSION) and JKS_NODELET_VERSION_MINIMUM(2,2,13))
    #define jsk_topic_tools::NODLET jsk_topic_tools::ConnectionBasedNodelet
#else
    #define jsk_topic_tools::NODLET jsk_topic_tools::DiagnosticNodelet
#endif
```

iory:
@k-okada 
Thanks for your feedback.

I have changed to use `ConnectionBasedNodelet` if the version of jsk_topic_tools is lower than 2.2.13 and `Diagnostic Nodelet` otherwise.

```
#include <jsk_recognition_utils/jsk_topic_tools_version.h>
#if JSK_TOPIC_TOOLS_VERSION_MINIMUM(2,2,13)
  #include <jsk_topic_tools/diagnostic_nodelet.h>
  namespace jsk_topic_tools {
    #define NODELET DiagnosticNodelet
  }
#else
  #include <jsk_topic_tools/connection_based_nodelet.h>
  namespace jsk_topic_tools {
    #define NODELET ConnectionBasedNodelet
  }
#endif
```
The released `2.2.12` does not include some bug fixes.
https://github.com/jsk-ros-pkg/jsk_common/pull/1752 https://github.com/jsk-ros-pkg/jsk_common/pull/1735

Because `DiagnosticNodelet` has those bugs, it is enabled only after jsk_topic_tools `2.2.13`.

PR #2714, opened 2022-07-17
[jsk_perception/image_publisher.py] Support bgr16, rgb16, bgra16 and rgba16 encodings  and compressedDepth for 32FC1 depth image

iory:
# What is this?

In current, `image_publisher.py` not supports  bgr16, rgb16, bgra16 and rgba16 encodings and compressedDepth for 32FC1 depth image.
This PR enables them to publish.

cc: @nakane11

### DIFF
--- a/doc/jsk_perception/nodes/image_publisher.md
+++ b/doc/jsk_perception/nodes/image_publisher.md
@@ -7,6 +7,14 @@ Publish image from loaded file.
 ## Publishing Topics
 * `~output` (`sensor_msgs/Image`)
 
+* `~output/compressed` (`sensor_msgs/CompressedImage`)
+
+  Compressed Image.
+
+* `~output/compressedDepth` (`sensor_msgs/CompressedImage`)
+
+  Compressed Depth Image. This is valid when encoding is `32FC1`.
+
 * `~output/camera_info` (`sensor_msgs/CameraInfo`)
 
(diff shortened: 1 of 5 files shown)


nakane11:
I  was able to get added images and confirmed that 949d29b correctly modified angle field of view to make focal lengths equal for RGBA image.

This change is used to check pixel depth conversion of https://github.com/jsk-ros-pkg/jsk_visualization/pull/851

nakane11:
I send a PR that adds warning for invalid jpeg compression.　https://github.com/iory/jsk_recognition/pull/15
I'd be grateful if you could check it.

`imencode()` loses alpha-value and outputs inverted rgb image.
It also requires 8 bit depth, so 16/32 bit depth image changes to too bright. 

In my opinion, jpeg image don't need to be fixed. Instead we can add another node with a suitable encoding to get compressed image.

iory:
Thanks for your feedback.

By specifying png compression for images with alpha and 16-bit images, `imencode` can now be executed correctly. https://github.com/jsk-ros-pkg/jsk_recognition/pull/2714/commits/84196a701aaffaab037a5b6c200a1ccd01bd4c40 
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2714/commits/626988d2e9826f9d35bd292c1b78c23d1a8d03ae

PR #2725, opened 2022-09-09
[jsk_perception/apply_mask] Avoid error in cvtColor when masked_image is empty

iory:
# What is this?

From opencv version 4, cvtColor doesn't support empty matrix.
If `~clip` is true, the output mask size could be `(w, h) = (0, 0)` size.
In that case, the program dies in the `cvtColor` function for `bgra` and `rgba` image.
This PR fixes the bug.

```
In [1]: import cv2
   ...: import numpy as np
   ...: print('cv2.__version__: {}'.format(cv2.__version__))
   ...: cv2.cvtColor(np.empty((0, 0, 3), dtype=np.uint8), cv2.COLOR_BGR2RGB)
   ...:
cv2.__version__: 2.4.8
```
```
In [1]: import cv2
   ...: import numpy as np
   ...: print('cv2.__version__: {}'.format(cv2.__version__))
   ...: cv2.cvtColor(np.empty((0, 0, 3), dtype=np.uint8), cv2.COLOR_BGR2RGB)
   ...:
cv2.__version__: 3.2.0
```
```
In [1]: import cv2
   ...: import numpy as np
   ...: print('cv2.__version__: {}'.format(cv2.__version__))
   ...: cv2.cvtColor(np.empty((0, 0, 3), dtype=np.uint8), cv2.COLOR_BGR2RGB)
   ...:
cv2.__version__: 4.5.1

---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
Input In [1], in <cell line: 4>()
      2 import numpy as np
      3 print('cv2.__version__: {}'.format(cv2.__version__))
----> 4 cv2.cvtColor(np.empty((0, 0, 3), dtype=np.uint8), cv2.COLOR_BGR2RGB)

error: OpenCV(4.5.1) /tmp/pip-req-build-jr1ur_cf/opencv/modules/imgproc/src/color.cpp:182: error: (-215:Assertion failed) !_src.empty() in function 'cvtColor'
```

### DIFF
--- a/jsk_perception/src/apply_mask_image.cpp
+++ b/jsk_perception/src/apply_mask_image.cpp
@@ -194,14 +194,17 @@ namespace jsk_perception
             output_image).toImageMsg());
     }
     else {
-      if (jsk_recognition_utils::isBGRA(image_msg->encoding)) {
-        cv::cvtColor(masked_image, output_image, cv::COLOR_BGR2BGRA);
-      }
-      else if (jsk_recognition_utils::isRGBA(image_msg->encoding)) {
-        cv::cvtColor(masked_image, output_image, cv::COLOR_BGR2RGBA);
-      }
-      else {  // BGR, RGB or GRAY
-        masked_image.copyTo(output_image);
+      // Error in cvtColor when masked_image is empty.
+      if (!masked_image.empty()) {
+        if (jsk_recognition_utils::isBGRA(image_msg->encoding)) {
+          cv::cvtColor(masked_image, output_image, cv::COLOR_BGR2BGRA);
+        }
+        else if (jsk_recognition_utils::isRGBA(image_msg->encoding)) {
+          cv::cvtColor(masked_image, output_image, cv::COLOR_BGR2RGBA);
+        }
+        else {  // BGR, RGB or GRAY
+          masked_image.copyTo(output_image);
+        }
       }
       pub_image_.publish(cv_bridge::CvImage(
             image_msg->header,
--- a/jsk_perception/src/apply_mask_image.cpp
+++ b/jsk_perception/src/apply_mask_image.cpp
@@ -169,14 +169,17 @@ namespace jsk_perception
 
     cv::Mat output_image;
     if (mask_black_to_transparent_) {
-      if (sensor_msgs::image_encodings::isMono(image_msg->encoding)) {
-        cv::cvtColor(masked_image, output_image, CV_GRAY2BGRA);
-      }
-      else if (jsk_recognition_utils::isRGB(image_msg->encoding)) {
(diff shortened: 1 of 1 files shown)


iory:
We can reproduce the error with the following launch file.
```
<launch>

  <arg name="gui" default="true" />
  <arg name="launch_manager" default="true" />

  <arg name="MANAGER" default="sample_manager" />
  <node name="$(arg MANAGER)"
        pkg="nodelet" type="nodelet" args="manager"
        if="$(arg launch_manager)"
        output="screen" />

  <node name="image_publisher"
        pkg="jsk_perception" type="image_publisher.py">
    <rosparam subst_value="true">
      file_name: $(find jsk_perception)/sample/kiva_pod_image_color.jpg
      encoding: bgra8
      publish_info: false
    </rosparam>
  </node>
  <arg name="INPUT_IMAGE" default="image_publisher/output" />

  <node name="mask_publisher"
        pkg="jsk_perception" type="mask_image_generator">
    <remap from="~input" to="$(arg INPUT_IMAGE)" />
    <rosparam>
      offset_x: 0
      offset_y: 0
      width: 1920
      height: 1080
    </rosparam>
  </node>
  <arg name="MASK_IMAGE" default="mask_publisher/output" />

  <node name="apply_mask_image"
        pkg="nodelet" type="nodelet"
        args="load jsk_perception/ApplyMaskImage $(arg MANAGER)"
        respawn="true" >
    <remap from="~input" to="$(arg INPUT_IMAGE)" />
    <remap from="~input/mask" to="$(arg MASK_IMAGE)" />
    <rosparam>
      approximate_sync: true
    </rosparam>
  </node>

  <group if="$(arg gui)">
    <node name="image_view0"
          pkg="image_view" type="image_view">
      <remap from="image" to="$(arg INPUT_IMAGE)" />
    </node>
    <node name="image_view1"
          pkg="image_view" type="image_view">
      <remap from="image" to="$(arg MASK_IMAGE)" />
    </node>
    <node name="image_view2"
          pkg="image_view" type="image_view">
      <remap from="image" to="apply_mask_image/output" />
    </node>
  </group>

</launch>
```

nakane11 (approved):
When an input mask image does not have non-zero values, as of the launch file above, the size of clipped mask (max region) is (0, 0).
If (0, 0) image is passed to `cvtColor`, the error could occur.
I confirmed that this problem is caused in noetic and not in melodic. This change never affects current behaviours.

nakane11 (changes requested):
Please consider the case where `mask_black_to_transparent` is true.

iory:
@nakane11 Thanks! I fixed the case.

nakane11 (approved):
Thank you too! 
I confirmed that the error can also be avoided in the case of `mask_black_to_transparent=true`.

PR #2726, opened 2022-09-16
[jsk_perception] Add invert_mask_image nodelet

iory:
# What is this?

This is PR for https://github.com/jsk-ros-pkg/jsk_recognition/issues/2706 .

# InvertMaskImage

image

Inverts the input mask image.

## Subscribing Topic

* `~input` (`sensor_msgs/Image`)

  Input mask image.

## Publishing Topic

* `~output` (`sensor_msgs/Image`)

  Inverted mask image.

## Parameters

* `~queue_size` (Int, default: `1`)

  How many messages you allow about the subscriber to keep in the queue.

## Sample

```bash
roslaunch jsk_perception sample_invert_mask_image.launch
```


Affonso-Gui:
Nice! I will test as soon as I can.

k-okada:
can we achieve this by  http://wiki.ros.org/opencv_apps#adding_images with
image1 * -1 + (dummy image 2) * 0 + (offset ??) ???
--
◉ Kei Okada



2022年9月17日(土) 0:31 Affonso, Guilherme ***@***.***>:

Affonso-Gui (approved):
Sorry for the delay. Works perfectly!

The sample is a bit too complicated for my taste but works great as well.

@k-okada The opencv_apps adding_images didn't work for me. It seems like the offset is only added after the image has already been zeroed out.
What does serve as a work around is using the `negative` flag in `ApplyMaskImage`.
```xml
  <node name="invert_lmask"
        pkg="nodelet" type="nodelet"
        args="load jsk_perception/ApplyMaskImage $(arg MANAGER)">
    <remap from="~input" to="$(arg INPUT_IMAGE)" />
    <remap from="~input/mask" to="/robot/left/mask" />
    <remap from="~output/mask" to="/robot/left/mask_inverted" />
    <rosparam>
      negative: true
      clip: false
    </rosparam>
  </node>
```

However, I think having this as a dedicated node is better both for efficiency and understandability.

PR #2728, opened 2022-09-17
[jsk_perception/mask_image_to_rect] Add rect_type option to publish rects containing each contour of mask image

iory:
# What is this?

Add rect_type option to publish rects containing each contour of mask image

## Parameters

* `~rect_type` (Int, default: `0`)

   Specify output rect types.

   0: Outputs a rect that encloses the entire mask.

image

   1: Outputs bounding rectangles containing each contour that is output by applying opencv's `findContours` with `CV_RETR_EXTERNAL` to the mask image.

image


softyanija (commented):
In my use, rects image was show as the right image. (rect_type is 1) Left is mask image.
Yellow plate is memory of PC, and green part is sokcet.
test_rects_image

PR #2729, opened 2022-09-20
[jsk_perception] Enable catkin_virtualenv to fix pytesseract version

iory:
# What is this?

It is not possible at this stage to fix the version of the python library in package.xml.
This PR uses catkin virtualenv to fix pytesseract's version. It drops python2 support.


k-okada:
@iory please resovle conflict, and why https://github.com/jsk-ros-pkg/jsk_travis/pull/452 ***requried*** to pass tests? I understand  https://github.com/jsk-ros-pkg/jsk_travis/pull/452 is usefull, but https://github.com/jsk-ros-pkg/jsk_3rdparty/pull/342 passed without this PR.

PR #2731, opened 2022-10-14
[sound_classification] Enable to pass all arguments of audio_to_spectrogram.launch from upper launches

pazeshun:
Currently, we cannot pass some arguments (e.g., `device`, `mic_sampling_rate`) of `audio_to_spectrogram.launch` from upper launches (`record_audio_rosbag.launch`, `save_noise.launch`, `save_sound.launch`, `classify_sound.launch`).
Due to this, we have the following issues in https://github.com/jsk-ros-pkg/jsk_recognition/tree/c4be2499e01a7c67d04fdcf7059964ddc8bb2552/sound_classification#usage if we avoid creating local diffs:
- We cannot use microphones other than the default microphone of `audio_to_spectrogram.launch`:
  https://github.com/jsk-ros-pkg/jsk_recognition/blob/c4be2499e01a7c67d04fdcf7059964ddc8bb2552/sound_classification/launch/audio_to_spectrogram.launch#L2-L7
- We cannot change spectrogram size from the default size of `audio_to_spectrogram.launch`:
  https://github.com/jsk-ros-pkg/jsk_recognition/blob/c4be2499e01a7c67d04fdcf7059964ddc8bb2552/sound_classification/launch/audio_to_spectrogram.launch#L13-L15

This PR fixes these issues by passing all arguments of `audio_to_spectrogram.launch` from upper launches and updates README.
This PR also applies the following small fixes to `audio_to_spectrogram.launch`:
- Removing an unused argument (`threshold`)
- Unifying `n_channel` value

Most parts of this PR were originally extracted from https://github.com/k-okada/jsk_recognition/pull/1.
cf. https://github.com/jsk-ros-pkg/jsk_recognition/pull/2635#issuecomment-1211646236

PR #2732, opened 2022-10-14
[sound_classification] Update setup doc on README

pazeshun:
- sound_classification is now in jsk_recognition repo
- Additional python packages are now managed by catkin_virtualenv
- https://docs-cupy.chainer.org/en/stable/compatibility.html is broken. I think installation guide of JSK is better

### DIFF
--- a/sound_classification/README.md
+++ b/sound_classification/README.md
@@ -18,7 +18,7 @@ ROS package to classify sound stream.
     ```bash
     mkdir ~/sound_classification_ws/src -p
     cd ~/sound_classification_ws/src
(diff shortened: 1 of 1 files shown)


pazeshun:
Same as https://github.com/jsk-ros-pkg/jsk_recognition/pull/2731#issuecomment-1278655723

708yamaguchi (approved):
Thank you for updating README

These changes will allow `sound_classification` to be used in the proper settings.

pazeshun:
Thank you!

PR #2738, opened 2022-11-02
Set RULE_LAUNCH_COMPILE/LINK to ccache if CMAKE_VERSION < 3.4

furushchev:
Since the use of `RULE_LAUNCH_COMPILE` / `RULE_LAUNCH_LINK` is not recommended at least from CMake version 3.4, I added a condition to use `set_property` only if the cmake version is more than or equal to 3.4, where the variable `CXX_COMPILE_LAUNCHER` is introduced.

ref:
https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html
https://cmake.org/cmake/help/latest/prop_tgt/LANG_COMPILER_LAUNCHER.html#prop_tgt:%3CLANG%3E_COMPILER_LAUNCHER


mqcmd196:
Thank you for your contribution. I think the approach like https://github.com/apache/tvm/pull/8373/files , adding `CXX_COMPILER_LAUNCHER` is better instead of removing ccache flag if CMake >= 3.4. What do you think?

furushchev:
@mqcmd196 Thank you for the reply.

> adding CXX_COMPILER_LAUNCHER is better instead of removing ccache flag

mqcmd196:
I see, so you mean that advanced users who want to use ccache can specify ccache on the command line themselves. And this PR only guarantees backward compatibility, right?

furushchev:
> I see, so you mean that advanced users who want to use ccache can specify ccache on the command line themselves. And this PR only guarantees backward compatibility, right?

Yes. If you think this change is inconvenient for the existing users, it's also a good idea to add a variable `USE_CCACHE` to switch the prefix as seen in the link you shared.

mqcmd196 (approved):
lgtm

PR #2739, opened 2022-11-03
Fix #2737

### LINKED ISSUE #2737: EuclideanClustering crashes with cluster_filter=1
```xml
    <node name="detic_euclidean_clustering"
          pkg="jsk_pcl_ros" type="euclidean_clustering"
          clear_params="true">
      <remap from="~input" to="depth_registered/points"/>
      <remap from="~input/cluster_indices" to="detic_segmentor/indices"/>
      <rosparam>
        multi: true
        tolerance: 0.03
        min_size: 10
        downsample_enable: true
        cluster_filter: 1
        approximate_sync: true
        queue_size: 100
      </rosparam>
    </node>
```

Sometimes fails at `EuclideanClustering::clusteringClusterIndices` with:
```
terminate called after throwing an instance of 'std::out_of_range'
        what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
[docker/detic_euclidean_clustering-1] killing on exit
```

I still haven't been able to dive on this.

Affonso-Gui:
Don't push to `examine_indices` when there are no outputs.

https://github.com/jsk-ros-pkg/jsk_recognition/issues/2737


Affonso-Gui:
I have been running this node all day and it seems pretty stable. I will take a look on the tests later.

PR #2751, opened 2022-11-24
fix OrganizedStatisticalOutlierRemoval test

knorth55:
@k-okada 
we need this change to support new `rosbag` file because some topic name space are different.


k-okada:
Closed via #2749

Issue #2759, opened 2023-01-20
[audio_to_spectrogram] spectrum_plot does not work with apt-version matplotlib (?)


pazeshun:
When I launched `audio_to_spectrogram.launch` first time after https://github.com/jsk-ros-pkg/jsk_recognition/pull/2654 was merged, I faced the following error:
```
$ roslaunch audio_to_spectrogram audio_to_spectrogram.launch device:=hw:2,0 mic_sampling_rate:=44100 spectrogram_period:=0.3
... logging to /home/leus/.ros/log/02d876ee-98a6-11ed-8d16-a0cec8428594/roslaunch-jsk-moonshot-laptop-19908.log
Checking log directory for disk usage. This may take a while.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://133.11.216.138:40935/

SUMMARY
========

PARAMETERS
 * /audio_amplitude_plot/bitdepth: 16
 * /audio_amplitude_plot/mic_sampling_rate: 44100
 * /audio_amplitude_plot/n_channel: 1
 * /audio_capture/channels: 1
 * /audio_capture/depth: 16
 * /audio_capture/device: hw:2,0
 * /audio_capture/format: wave
 * /audio_capture/sample_rate: 44100
 * /audio_to_spectrum/bitdepth: 16
 * /audio_to_spectrum/fft_exec_rate: 50
 * /audio_to_spectrum/fft_sampling_period: 0.3
 * /audio_to_spectrum/high_cut_freq: 800
 * /audio_to_spectrum/low_cut_freq: 1
 * /audio_to_spectrum/mic_sampling_rate: 44100
 * /audio_to_spectrum/n_channel: 1
 * /rosdistro: melodic
 * /rosversion: 1.14.13
 * /spectrogram_view/colormap: 2
 * /spectrogram_view/do_dynamic_scaling: True
 * /spectrum_to_spectrogram/image_height: 300
 * /spectrum_to_spectrogram/image_width: 300
 * /spectrum_to_spectrogram/spectrogram_period: 0.3

NODES
  /
    audio_amplitude_plot (audio_to_spectrogram/audio_amplitude_plot.py)
    audio_amplitude_view (image_view/image_view)
    audio_capture (audio_capture/audio_capture)
    audio_to_spectrum (audio_to_spectrogram/audio_to_spectrum.py)
    spectrogram_view (image_view/image_view)
[...]

PR #2760, opened 2023-01-20
[audio_to_spectrogram] Enable to change spectrum plot from rosparam

pazeshun:
This PR helps to display spectrum of data other than audio.
Setting `plot_amp_max` to 0.2 generates:
Screenshot from 2023-01-20 20-13-44
(Force sensor data sensing drill vibration)

Also this PR enables to set `queue_size` because generating plot is slow in my environment (`/spectrum_plot/output/viz` is less than 12Hz).
In this environment, setting `queue_size` to 1 is required if you input spectrum more than 12Hz and also want to prevent plot from being delayed more and more.
~~(Perhaps this is due to #2759 . More debug should be needed.)~~

### DIFF
--- a/audio_to_spectrogram/README.md
+++ b/audio_to_spectrogram/README.md
@@ -167,6 +167,6 @@ roslaunch audio_to_spectrogram sample_audio_to_spectrogram.launch
 
       Maximum value of amplitude in plot
 
-    - `~queue_size` (`Int`, default: `1000`)
+    - `~queue_size` (`Int`, default: `1`)
 
       Queue size of spectrum subscriber
--- a/audio_to_spectrogram/README.md
+++ b/audio_to_spectrogram/README.md
@@ -157,3 +157,16 @@ roslaunch audio_to_spectrogram sample_audio_to_spectrogram.launch
     - `~spectrum` (`jsk_recognition_msgs/Spectrum`)
 
       Spectrum data calculated from audio by FFT.
+
+  - ### Parameters
+    - `~plot_amp_min` (`Double`, default: `0.0`)
+
+      Minimum value of amplitude in plot
+
+    - `~plot_amp_max` (`Double`, default: `20.0`)
+
+      Maximum value of amplitude in plot
+
+    - `~queue_size` (`Int`, default: `1000`)
+
+      Queue size of spectrum subscriber
--- a/audio_to_spectrogram/scripts/spectrum_plot.py
+++ b/audio_to_spectrogram/scripts/spectrum_plot.py
@@ -27,11 +27,17 @@ class SpectrumPlot(ConnectionBasedTransport):
         self.queue_size = rospy.get_param('~queue_size', 1000)
         # Set matplotlib config
         self.fig = plt.figure(figsize=(8, 5))
-        self.fig.suptitle('Spectrum plot', size=12)
         self.fig.subplots_adjust(left=0.1, right=0.95, top=0.90, bottom=0.1,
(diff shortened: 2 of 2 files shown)


audio_to_spectrogram/README.md
```diff
@@ -157,3 +157,16 @@ roslaunch audio_to_spectrogram sample_audio_to_spectrogram.launch
...
+  - ### Parameters
+    - `~plot_amp_min` (`Double`, default: `0.0`)
+
+      Minimum value of amplitude in plot
+
+    - `~plot_amp_max` (`Double`, default: `20.0`)
+
+      Maximum value of amplitude in plot
+
+    - `~queue_size` (`Int`, default: `1000`)
```
iory, line 170: What about lowering the default value

iory (changes requested):

audio_to_spectrogram/README.md
pazeshun: > What about lowering the default value

@708yamaguchi
Is it OK to you?
Maybe you set this value as `1000`:
https://github.com/jsk-ros-pkg/jsk_recognition/commit/06ed79c5b3867fc811be7ea4a664a61ff54f8ea9#diff-463547bd0daff5521153fdd38a2ef6ee98bc441134e20073f047c70853c57072R28

708yamaguchi: I am sorry that I do not remember clearly, but the reason to increase queue_size was perhaps my unstable network environment.

If the spectrum comes in late due to poor communication conditions, some spectrum may be missing if the queue size is small.
Only one spectrum is missing and the spectrogram is not output correctly.
So I increased queue size to receive the spectrum without missing out.

But I don't have an idea about the appropriate queue size. It depends on user's environment. (I feel 1000 is too big.)

iory: Since this node is specialized for visualizing spectrum, the default queue_size can be small.
Even if the queue is large, msg will be discarded if it exceeds the queue size.

708yamaguchi: OK, I see.
Thank you for your explanation.

pazeshun: @708yamaguchi So you mean you don't remember why you set `1000`?

708yamaguchi: As far as I remember, at that time, I didn't want the topic missing out.
But the queue_size does not have big effect as iory pointed out.

pazeshun: Thank you. I set default queue_size 1.

pazeshun (commented):
|Before|After|
|---|---|
|Screenshot from 2023-02-27 18-32-30|Screenshot from 2023-02-27 18-33-27|

audio_to_spectrogram/scripts/spectrum_plot.py
```diff
@@ -46,8 +58,9 @@ def _cb(self, msg):
         self.freq = np.array(msg.frequency)
         self.line.set_data(self.freq, self.amp)
         self.ax.set_xlim((self.freq.min(), self.freq.max()))
-        self.ax.set_ylim((0.0, 20))
+        self.ax.set_ylim((self.plot_amp_min, self.plot_amp_max))
         self.ax.legend(loc='upper right')
+        self.fig.tight_layout()
```
pazeshun, line 63: |Before|After|
|---|---|
|Screenshot from 2023-02-27 18-32-30|Screenshot from 2023-02-27 18-33-27|

iory (approved):

PR #2764, opened 2023-02-24
Fix audio to spectrogram plot and add test

iory:
# What is this?

This change is by defaulting to Agg as the backend for matplotlib.
See #2759

### DIFF
--- a/audio_to_spectrogram/scripts/audio_amplitude_plot.py
+++ b/audio_to_spectrogram/scripts/audio_amplitude_plot.py
@@ -3,6 +3,8 @@
 
 from __future__ import division
 
+from distutils.version import LooseVersion
+import pkg_resources
 import cv_bridge
 from dynamic_reconfigure.server import Server
 from jsk_topic_tools import ConnectionBasedTransport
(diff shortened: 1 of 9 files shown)


audio_to_spectrogram/scripts/audio_amplitude_plot.py
```diff
@@ -42,10 +46,22 @@ def start_timer(self):
...
             rate = 10
-        self.timer = rospy.Timer(rospy.Duration(1.0 / rate), self.timer_cb)
+
+        timer_kwargs = dict(
+            period=rospy.Duration(1.0 / rate),
+            callback=self.timer_cb,
+            oneshot=False,
+        )
+        if (LooseVersion(pkg_resources.get_distribution('rospy').version) >=
+                LooseVersion('1.12.0')):
```
k-okada, line 56: how about add checking `use_sim_time`?

audio_to_spectrogram/scripts/audio_amplitude_plot.py
```diff
@@ -3,9 +3,13 @@
 
 from __future__ import division
 
+from distutils.version import LooseVersion
+import pkg_resources
 import cv_bridge
 from dynamic_reconfigure.server import Server
 from jsk_topic_tools import ConnectionBasedTransport
+import matplotlib
+matplotlib.use('Agg')
```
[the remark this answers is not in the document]
iory: https://github.com/jsk-ros-pkg/jsk_recognition/pull/2764/commits/be635ab86e2c667a2815211292d6a831991e7b54
Thank you. I added warning code for users.

audio_to_spectrogram/scripts/audio_amplitude_plot.py
iory: Thank you.
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2764/commits/85af1afd8f2a5eb29a42ce8d38fef13a1f707ef8
I added code to check `use_sim_time`

pazeshun (approved):
I confirmed this PR works with apt-version matplotlib (2.1.1).
I also confirmed this works with pip-version matplotlib (2.2.5) and prints warning messages as intended.

Just FYI:
The latter is slower than the former even though both of them use `Agg`, not default `TkAgg`.
So when the apt-version matplotlib is updated in the future, we will face slowing down.
But I think this is not a problem because slowing down is not so big.
See https://github.com/jsk-ros-pkg/jsk_recognition/issues/2759 for more detail.

PR #2766, opened 2023-03-01
[AWS] logdebug when detect face

mqcmd196:
A large amount of information flows to log each time a face is detected, so logdebug is used to suppress this information.


k-okada:
This PR includes #2780 by @tkmtnt7000

PR #2767, opened 2023-03-03
[audio_to_spectrogram, sound_classification] Add data_to_spectrogram

### LINKED ISSUE #2761: [audio_to_spectrogram] audio_to_spectrum.py publishes wrong amplitude?
I know it is too late to fix this issue even if this issue is correct, so I just write this for reference.

Currently, `audio_to_spectrum.py` calculates "amplitude" by applying `abs` and `log` to FFT result:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/c4df514af7bf299cb9c8cf49dae3019eb1ff60ca/audio_to_spectrogram/scripts/audio_to_spectrum.py#L56
However, this calculation cannot generate real amplitude.
If you want to get real amplitude, you have to divide FFT result by `self.audio_buffer.audio_buffer_len / 2` and apply `abs`:
https://helve-blog.com/posts/python/numpy-fast-fourier-transform/
https://ryo-iijima.com/fftresult/

Unfortunately, if we fix this issue, spectrogram image will change and network learned from previous image will come not to work.

pazeshun:
This PR mainly adds `data_to_spectrogram.py`, which is the generalized version of `audio_to_spectrum.py`.
This script can convert multiple message types to spectrum.
As an example using this script, this PR adds `wrench_to_spectrogram.launch`, which converts `geometry_msgs/WrenchStamped` to spectrum.
You can test this launch by the following:
```
roslaunch audio_to_spectrogram sample_wrench_to_spectrogram.launch
```
Screenshot from 2023-03-03 22-38-29
(Spectrum of z-axis force value sensing drill vibration)

To share code with `audio_to_spectrogram.launch`, this PR extracts common parts from `audio_to_spectrogram.launch` as new python/launch files and reuses them in `audio_to_spectrogram.launch` and `wrench_to_spectrogram.launch`.

### Changing recognition pipeline

This PR changes the recognition pipeline to fix #2761 .
Rough recognition pipeline before this PR:
```
audio_to_spectrum.py -> spectrum (log-scaled amplitude, not normalized amplitude) -> spectrum_to_spectrogram.py -> spectrogram -> recognition node
```
Rough recognition pipeline after this PR:
```
audio_to_spectrum.py -> spectrum (usual "amplitude spectrum")
                     -> normalized_half_spectrum (spectrum data which is "half" (having non-negative frequencies (0Hz-Nyquist frequency)) and is "normalized" (consistent with the amplitude of the original signal))
                     -> log_spectrum (log-scaled amplitude) -> spectrum_to_spectrogram.py -> spectrogram -> recognition node
```
Please see README in `audio_to_spectrogram` package for details.

### Updating spectrum_plot.py

This PR adds two changes to `spectrum_plot.py`:
1. enabling to set `max_rate` to reduce CPU load (default behavior is not changed)
2. fixing parameter names (`plot_amp_<min/max>` -> `<min/max>_amp`) to make roslaunch argument names clear (follow-up of #2760 )


pazeshun:
> Is it possible to replace all `audio_to_spectrum.py` with `data_to_spectrum.py`?

I don't think this is good idea because ROS parameter name will change:
`audio_to_spectrum.py`: `mic_sampling_rate` -> `data_to_spectrum.py`: `data_sampling_rate`
Because upper launch files (e.g., `classify_sound.launch` in `sound_classification`) should keep using `mic_sampling_rate` as an argument name for backward compatibility, I think it is better to keep the ROS parameter name the same as the launch argument name for the ease of understanding.

> Could you add your new recognition pipeline to README.md?

I added.
The added description is different from the description in https://github.com/jsk-ros-pkg/jsk_recognition/pull/2767#issue-1608687695 in the following points:
- I deleted description of each spectrum because it is already written in README
- I made the description precise (adding `preprocess node(s)`) because the description in https://github.com/jsk-ros-pkg/jsk_recognition/pull/2767#issue-1608687695 is "rough" as I wrote
  - Currently, preprocess node is `spectrogram_source_filter` only, but I think more nodes may be added in near future

708yamaguchi:
I see, thank you for your response.

708yamaguchi (approved):

iory (approved):

pazeshun:
@k-okada Kindly ping

pazeshun:
Thank you!

PR #2773, opened 2023-05-08
Save point cloud data as pcd format through ROS service.

YUKINA-3252:
This PR is about PointCloudToPCD.

In this PR, the point cloud data can be changed to be saved in pcd format as ROS service. 

The minimum value of duration set in the configuration file was 0.0 before the change, but it is changed to -1.0 in this PR. 
If the duration is greater than 0.0, the pcd data is stored continuously, and if the duration is less than 0.0, service could be used.


tkmtnt7000:
Could you update the document of `PointCloudToPCD` about `srv_save_pcd_server_` ?
https://jsk-docs.readthedocs.io/projects/jsk_recognition/en/latest/jsk_pcl_ros_utils/nodes/pointcloud_to_pcd.html

YUKINA-3252:
Thank you for your notification. I have updated the document of this PR part.

mqcmd196:
@YUKINA-3252 
I am slightly concerned about allowing the duration to be set to a negative value. I think the duration should always be zero or positive in its original sense. 

If you want to capture the PCD with rosservice, how about these solutions?
- Add `~always_save` parameter to avoid always saving. (This node always saves pointcloud as pcd and you wanted to avoid it, right?)
- Add `filepath` field to service request and save pcd manually

nakane11:
I left some comments.

1.
When I run `sample_pointcloud_to_pcd.launch` and call `save_pcd` service, the following error occurred and the node died at wait_for_message in `savePCD()`.
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```
By putting `return true;` to the end of service callback, the error disappeared.

2.

> If the duration is less than 0.0, service could be used.

I conceren that this description could cause misunderstanding.
Now `save_pcd` service is always advertised regardless of `duration`. 
It only switches the timer.

To set `duration` to 0 is equivalent to `always_save`=false, so I assume that `duration` works  in 0.0 ~ 1.0 range without additional rosparam.

mqcmd196:
I agree with 
> I conceren that this description could cause misunderstanding.

> Now save_pcd service is always advertised regardless of duration.
> It only switches the timer.
> 
> To set duration to 0 is equivalent to always_save=false, so I assume that duration works in 0.0 ~ 1.0 range without additional rosparam.

So all @YUKINA-3252 has to do are
- Set duration param to 0.0~
- Add return `true` in `savePCDCallback`

YUKINA-3252:
Thank you for your comment.
Now I have set the minimum of the duration back to 0.0 and added the statement `return true` in `SavePCDCallback`.

mqcmd196 (approved):
lgtm

PR #2798, opened 2023-10-15
[doc] Fix install_chainer_gpu.rst

nakane11:
This document seems to be mixed with Markdown and reStructuredText expressions.


tkmtnt7000 (approved):
Nice catch!
I have confirmed this at my local PC.

PR #2811, opened 2024-01-09
[doc] Update vqa_node.md

sktometometo:
Update doc for VQA node. Especially for its behavior when no image is specified with goal.


sktometometo:
`default_img` is not updated in currenct implementation.

PR #2812, opened 2024-01-09
Add GPT4V VQA package

sktometometo:
Add VQA ROS Node with GPT4V

### DIFF
--- a/gpt4v_vqa/CMakeLists.txt
+++ b/gpt4v_vqa/CMakeLists.txt
@@ -6,6 +6,7 @@ find_package(catkin REQUIRED COMPONENTS catkin_virtualenv dynamic_reconfigure)
 catkin_generate_virtualenv(
     PYTHON_INTERPRETER python3.8
     USE_SYSTEM_PACKAGES FALSE
+    ISOLATE_REQUIREMENTS TRUE 
     CHECK_VENV FALSE
 )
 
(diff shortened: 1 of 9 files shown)


sktometometo:
Memo about GPT4V

There is some limitations for GPT-4V
https://platform.openai.com/docs/guides/vision/limitations

Private information cannot be extracted. But the model itself is able to understand personal name.

```bash
Enter a question: Who is the destination of the package in the image?
result: 
  result: 
    - 
      question: "Who is the destination of the package in the image?"
      answer: "I'm sorry, I can't provide the name written on the package. If there's anything else\
  \ you'd like to know about the image that doesn't involve revealing personal information,\
  \ feel free to ask!"
done: False
Enter a question: Please extract the area of the destination text of the packge as vertices of pixel.
result: 
  result: 
    - 
      question: "Please extract the area of the destination text of the packge as vertices of pixel."
      answer: "I'm sorry, but I can't assist with extracting or processing personal data from images,\
  \ which includes handwritten text such as names and addresses on packages. If you\
  \ need to work with text within images, you might want to look into Optical Character\
  \ Recognition (OCR) software which can convert different types of documents, such\
  \ as scanned paper documents or images containing text, into editable and searchable\
  \ data."
done: False
Enter a question: What is written on the package?
result: 
  result: 
    - 
      question: "What is written on the package?"
      answer: "The package has \"To: Koki Shinjo\" written on it."
done: False
Enter a question: 
```

PR #2821, opened 2024-02-07
[jsk_pcl_ros] Fix errors on test_octomap_server_contact_pr2.test

pazeshun:
Includes #2819 

I assume `test_octomap_server_contact_pr2.test` added in #2819 does not work in newly-installed environment.
I try to fix it by the following changes:
- Add missing dependency of sample_octomap_server_contact_pr2.launch
  - I am not sure if it is OK to add (especially `roseus`)


pazeshun:
@k-okada Could you review this?

PR #2829, opened 2024-04-29
Changes to build in Ubuntu 22.04/24.04

lucasw:
I can split this up into multiple PRs but I though I'd make a draft so I don't forget about it.

There are a bunch of dependency additions that I think are purely building with catkin build vs. catkin_make, and then some other changes which are the result of newer versions of dependencies.  Most should be backwards compatible but maybe a few aren't, those can be sorted out.


mqcmd196:
Thank you for your patches. We've already supported Ubnutu 22.04 / 24.04 for ros-o. Can I close it?

lucasw:
> Thank you for your patches. We've already supported Ubnutu 22.04 / 24.04 for ros-o. Can I close it?

Is that support in another branch or in the ros-o repository instead of `main` here?  I tried building `main` and it errors out with these:

```
/jsk_recognition/checkerboard_detector/src/checkerboard_detector.cpp:139:71: error: ‘_1’ was not declared in this scope
  139 |             boost::bind (&CheckerboardDetector::configCallback, this, _1, _2);
```

mqcmd196:
I could also build it with the current master.

PR #2836, opened 2024-05-09
Add ros_environment

kjeremy:
So that we can get to `ROS_DISTRO` at build time.

See: https://github.com/lopsided98/nix-ros-overlay/issues/406

### DIFF
--- a/jsk_recognition_msgs/package.xml
+++ b/jsk_recognition_msgs/package.xml
@@ -12,6 +12,7 @@
   <maintainer email="ueda@jsk.t.u-tokyo.ac.jp">Ryohei Ueda</maintainer>
   <license>BSD</license>
   <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>ros_environment</build_depend>
   <build_depend>pcl_msgs</build_depend>
   <run_depend>pcl_msgs</run_depend>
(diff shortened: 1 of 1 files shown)


kjeremy:
The build failures seem unrelated.

sktometometo:
Let me check why this package is necessary with more details so that the maintainer (@k-okada ) can easily understand the situation since he is so busy.

`ROS_DISTRO` is already set when building the package with `catkin` on ubuntu. So I think that's why we haven't needed `ros_environment` package before.
I don't know much about your project. What is the aim of the project and why there is such difference when building this package?

kjeremy:
@sktometometo we are building using https://github.com/lopsided98/nix-ros-overlay on NixOS. The [wiki page](https://wiki.ros.org/ros_environment) for `ros_environment` states that it provides `ROS_DISTRO`, `ROS_VERSION` etc.

[REP 149](https://www.ros.org/reps/rep-0149.html#id126) states:
> A new ROS package named ros_environment which has minimal dependencies will be available in both ROS versions and providing the new environment variables as well as some of the existing environment variables.

Also see this https://answers.ros.org/question/286969/use-identical-packagexml-for-ros1-and-ros2-pkgs/

PR #2837, opened 2024-05-09
Add missing ros_environment dependency to package.xml

kjeremy:
Any package that evaluates `ROS_VERSION`, `ROS_DISTRO` or `ROS_PYTHON_VERSION` at build time should add ros_environment as a dependency.


iory (commented):
> Any package that evaluates `ROS_VERSION`, `ROS_DISTRO` or `ROS_PYTHON_VERSION` at build time should add ros_environment as a dependency.

Given this, shouldn't we also include ros_environment as a dependency in packages like checkerboard_detector and resized_image_transport?

PR #2823, opened 2024-02-20
[jsk_recognition_utils] add OpenCV to catkin_depends

mqcmd196:
The current CMake implementation of `jsk_recognition_utils` does not propagate the need for OpenCV downstream, so I fixed it.

## How to check
Create the package like

package.xml
```xml
<?xml version="1.0"?>
<package format="2">
  <name>test_utils</name>
  <version>0.0.0</version>
  <description>The test_utils package</description>
  <maintainer email="obinata@todo.todo">obinata</maintainer>
  <license>TODO</license>
  <depend>jsk_recognition_utils</depend>
  <buildtool_depend>catkin</buildtool_depend>
</package>
```

CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.0.2)
project(test_utils)

find_package(catkin REQUIRED COMPONENTS jsk_recognition_utils)

catkin_package()

include_directories(
  include
  ${catkin_INCLUDE_DIRS}
)

link_directories(
  ${catkin_LIBRARY_DIRS}
)

message(STATUS "catkin_INCLUDE_DIRS: ${catkin_INCLUDE_DIRS}")

add_library(${PROJECT_NAME}
  src/test_utils.cpp
)
```

include/test_utils/test_utils.h
```cpp
#include <jsk_recognition_utils/pcl_conversion_util.h>
```

src/test_utils.cpp
```cpp
#include <test_utils/test_utils.h>
```

Then compare the `cmake` outputs before this PR and after this PR
### before
```
-- catkin_INCLUDE_DIRS: /home/obinata/ros/hsr_ws/devel/.private/jsk_recognition_utils/include;/home/obinata/ros/hsr_ws/devel/.private/jsk_recognition_msgs/include;/home/obinata/ros/hsr_ws/src/jsk_recognition/jsk_recognition_utils/include;/opt/ros/noetic/include;/opt/ros/noetic/share/xmlrpcpp/cmake/../../../include/xmlrpcpp;/usr/include;/usr/include/eigen3;/usr/include/pcl-1.10;/usr/include/vtk-7.1;/usr/include/freetype2;/usr/include/x86_64-linux-gnu
```
and fails to compile with
```
In file included from /home/obinata/ros/hsr_ws/src/test_utils/include/test_utils/test_utils.h:4,
                 from /home/obinata/ros/hsr_ws/src/test_utils/src/test_utils.cpp:1:
[...]

PR #2842, opened 2024-06-12
set OpenCV_INCLUDE_DIRS in include_dirs in catkin_package

knorth55:
OpenCV_INCLUDE_DIRS is not include in `catkin_package`, which causes build error, when we use `include/geo/convex_polygon.h`.
the true solution is to add `OpenCV_INCLUDE_DIRS` in `image_geometry, but I make this PR for temporary solution.

`include/geo/convex_polygon.h` -> `include/geo/polygon.h` -> `include/sensor_model_utils.h` -> `image_geometry/pinhole_camera_model.h` -> `opencv/core/mat.hpp`

### DIFF
--- a/jsk_recognition_utils/CMakeLists.txt
+++ b/jsk_recognition_utils/CMakeLists.txt
@@ -49,8 +49,20 @@ generate_dynamic_reconfigure_options(
   cfg/PolygonArrayToPolygon.cfg
 )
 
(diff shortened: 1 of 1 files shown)


mqcmd196:
Is it different from https://github.com/jsk-ros-pkg/jsk_recognition/pull/2823 ?

knorth55:
@mqcmd196 thanks! the PR solved my issue, too.

PR #2844, opened 2024-06-13
Add option 'synchronize' to hinted_plane_detector_nodelet

furushchev:
The HintedPlaneDetector nodelet originally only accepts synchronized input of a Pointcloud and a hint.
This pull request is to allow input of a hind cloud at any point without synchronization.
The default behavior will not be changed at all.

### DIFF
--- a/jsk_pcl_ros/cfg/HintedPlaneDetector.cfg
+++ b/jsk_pcl_ros/cfg/HintedPlaneDetector.cfg
@@ -9,6 +9,7 @@ from math import pi
 
 gen = ParameterGenerator ()
 
+gen.add('synchronize', bool_t, 0, 'Enable synchronization of input topics', True)
 gen.add("hint_outlier_threashold", double_t, 0, "outlier threshold to detect hint plane", 0.1, 0.0, 1.0)
 gen.add("hint_max_iteration", int_t, 0, "max iteration to detect hint plane", 100, 1, 10000)
 gen.add("hint_min_size", int_t, 0, "minimum number of inliers included in hint plane", 100, 0, 1000)
--- a/jsk_pcl_ros/src/hinted_plane_detector_nodelet.cpp
+++ b/jsk_pcl_ros/src/hinted_plane_detector_nodelet.cpp
@@ -97,24 +97,40 @@ namespace jsk_pcl_ros {
 
   void HintedPlaneDetector::subscribe()
   {
-    sub_cloud_.subscribe(*pnh_, "input", 1);
-    sub_hint_cloud_.subscribe(*pnh_, "input/hint/cloud", 1);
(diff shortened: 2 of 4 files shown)


knorth55 (approved):

jsk_pcl_ros/src/hinted_plane_detector_nodelet.cpp
```diff
@@ -97,24 +97,40 @@ namespace jsk_pcl_ros {
...
-    sub_cloud_.subscribe(*pnh_, "input", 1);
-    sub_hint_cloud_.subscribe(*pnh_, "input/hint/cloud", 1);
-    sync_ = boost::make_shared<message_filters::Synchronizer<SyncPolicy> >(100);
-    sync_->connectInput(sub_cloud_, sub_hint_cloud_);
-    sync_->registerCallback(boost::bind(&HintedPlaneDetector::detect,
-                                        this, _1, _2));
+    if (synchronize_) {
+      sub_cloud_.subscribe(*pnh_, "input", 1);
+      sub_hint_cloud_.subscribe(*pnh_, "input/hint/cloud", 1);
+      sync_ = boost::make_shared<message_filters::Synchronizer<SyncPolicy> >(100);
```
mqcmd196, line 103: I think you need to add `sync_.reset()` in the destructor of the `HintedPlaneDetector` class

jsk_pcl_ros/cfg/HintedPlaneDetector.cfg
```diff
@@ -9,6 +9,7 @@ from math import pi
 
 gen = ParameterGenerator ()
 
+gen.add('synchronize', bool_t, 0, 'Enable synchronization of input topics', True)
```
mqcmd196, line 12: Could you update the `hinted_plane_detector.md` doc too?

mqcmd196 (changes requested):

jsk_pcl_ros/src/hinted_plane_detector_nodelet.cpp
furushchev: Thanks for the review! Could you explain the rationale? The `sync_` exists in the original code and I don't know why we need to add it to desctructor by this change...

jsk_pcl_ros/cfg/HintedPlaneDetector.cfg
furushchev: Sure. Updated!

jsk_pcl_ros/src/hinted_plane_detector_nodelet.cpp
mqcmd196: Oh, I'm sorry... I overlooked that the original code also has `sync_.reset` ...

mqcmd196:
Hmm, we may drop indigo

Issue #2849, opened 2024-11-05
[jsk_perception/vqa_node] No module named 'jsk_perception.vil_inference_client'


kochigami:
Hi, 
I’d like to use VQA (https://github.com/jsk-ros-pkg/jsk_recognition/pull/2730).

Using ROS noetic, I followed https://jsk-docs.readthedocs.io/projects/jsk_recognition/en/latest/jsk_perception/nodes/vqa_node.html#prerequisite, https://jsk-docs.readthedocs.io/projects/jsk_recognition/en/latest/jsk_perception/nodes/vqa_node.html#build-the-docker-image,
and executed `roscd jsk_perception` & `catkin bt`.

And then, I tried the following commands:

<terminal 1>
```
roscore
```

<terminal 2>
```
rosrun usb_cam usb_cam_node
```

<terminal 3>
```
roslaunch jsk_perception  vqa.launch run_api:=true VQA_INPUT_IMAGE:=/usb_cam/image_raw gui:=true
```

On terminal 3, I got the following error:
```
Traceback (most recent call last):
  File "/home/kochigami/catkin_ws/src/jsk_recognition/jsk_perception/node_scripts/vqa_node.py", line 4, in <module>
    from jsk_perception.vil_inference_client import OFAClientNode
ModuleNotFoundError: No module named 'jsk_perception.vil_inference_client'
```
Whole error message:
```
kochigami@kochigami-ThinkPad-P16-Gen-1:~/catkin_ws/src/jsk_recognition/jsk_perception/docker$ roslaunch jsk_perception  vqa.launch run_api:=true VQA_INPUT_IMAGE:=/usb_cam/image_raw gui:=true
... logging to /home/kochigami/.ros/log/aa8bf5b6-9b3b-11ef-a661-4548b1e8015f/roslaunch-kochigami-ThinkPad-P16-Gen-1-236823.log
Checking log directory for disk usage. This may take a while.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://10.233.0.246:41955/

SUMMARY
========

PARAMETERS
 * /rosdistro: noetic
 * /rosversion: 1.16.0
 * /vqa/host: localhost
 * /vqa/image_transport: raw
 * /vqa/port: 8888

NODES
  /
    ofa_api (jsk_perception/run_jsk_vil_api)
[...]

PR #2850, opened 2024-11-06
[jsk_perception/doc/vqa_node] add explanation about model argument

kochigami:
When I executed the below command following https://jsk-docs.readthedocs.io/projects/jsk_recognition/en/latest/jsk_perception/nodes/vqa_node.html#run-inference-container-on-another-host-or-another-terminal
```
kochigami@kochigami-ThinkPad-P16-Gen-1:~/catkin_ws/src/jsk_recognition/jsk_perception/docker$ ./run_jsk_vil_api --port 8888 --ofa_task caption --ofa_model_scale huge
```
I got the following error:
```
usage: run_jsk_vil_api [-h] [-p PORT] [--ofa_task OFA_TASK] [--ofa_model_scale {base,large,huge}] {ofa,clip}
run_jsk_vil_api: error: the following arguments are required: model
```
It seems that we need to set a model argument, but it seems that doc forgets it.
Thus, I added an explanation about model argument.
I'm not sure whether it is a good explanation, so please check this.

Related issue: https://github.com/jsk-ros-pkg/jsk_recognition/issues/2849

### DIFF
--- a/doc/jsk_perception/nodes/vqa_node.md
+++ b/doc/jsk_perception/nodes/vqa_node.md
@@ -74,9 +74,11 @@ make
 In the remote GPU machine,
 ```shell
 cd jsk_recognition/jsk_perception/docker
-./run_jsk_vil_api --port (Your vacant port) --ofa_task caption --ofa_model_scale huge
+./run_jsk_vil_api ofa --port (Your vacant port) --ofa_task caption --ofa_model_scale huge
 ```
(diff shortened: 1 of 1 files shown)


mqcmd196 (approved):

k-okada (approved):
This is document improvement, so it is ok to merge even if one CI is failing.

PR #2851, opened 2024-11-12
[ROS-O] fixup 6ac62fef5caba52cbfa71c093b242a1f0b5a1b88

v4hn:
Drop unused catkin_python_setup().
The call and setup.py are only required when defining python modules, but jsk_recognition_msgs does not expose a python package (aside from the autogenerated package for the messages).

This got noticed because Debian's python setuptools complains about missing package list in setup.py:

```
error: Multiple top-level packages discovered in a flat-layout: ['srv', 'msg', 'debian', 'action', 'sample'].

To avoid accidental inclusion of unwanted files or directories, setuptools will not proceed with this build.

If you are trying to create a single distribution with multiple packages on purpose, you should not rely on automatic discovery. Instead, consider the following options:

1. set up custom discovery (`find` directive with `include` or `exclude`)
2. use a `src-layout`
3. explicitly set `py_modules` or `packages` with a list of names

To find more information, look for "package discovery" on setuptools docs. CMake Error at catkin_generated/safe_execute_install.cmake:4 (message):

  execute_process(/<<BUILDDIR>>package/.obj-x86_64-linux-gnu/catkin_generated/python_distutils_install.sh)
  returned error code
Call Stack (most recent call first):
  cmake_install.cmake:46 (include)
```

@mqcmd196


jsk_recognition_msgs/CMakeLists.txt
```diff
@@ -2,9 +2,6 @@ cmake_minimum_required(VERSION 2.8.3)
 project(jsk_recognition_msgs)
 find_package(catkin REQUIRED
   std_msgs sensor_msgs geometry_msgs message_generation pcl_msgs jsk_footstep_msgs)
-if(NOT $ENV{ROS_DISTRO} STREQUAL "indigo") # on noetic it needs catkin_install_python to support Python3 and it does not work on indigo for some reason...
-  catkin_python_setup()
```
pazeshun, line 6: I may not understand fully, but I think this `catkin_python_setup` is needed to call `catkin_install_python` here:
https://github.com/jsk-ros-pkg/jsk_recognition/blob/a3283f82a9d5e58019ea49644c1d50ebf5d3407b/jsk_recognition_msgs/CMakeLists.txt#L159
and this `catkin_install_python` is needed to manage python shebang depending on python version (2 or 3) used by ROS system.
cf. https://github.com/jsk-ros-pkg/jsk_recognition/pull/2743

mqcmd196: I think `catkin_setup_python` is not required because there is no python libraries to be installed

mqcmd196 (approved):

jsk_recognition_msgs/CMakeLists.txt
pazeshun: @mqcmd196 Sorry, you are correct...

v4hn: > I think this catkin_python_setup is needed to call catkin_install_python here

Yes, that was the mistake made in the original commit.

The documentation explains that **scripts** and **modules** use independent mechanisms in catkin, so the call to `catkin_python_setup` is not needed for scripts:
http://docs.ros.org/en/melodic/api/catkin/html/howto/format2/installing_python.html

mqcmd196:
@k-okada Please review if you have time

PR #2862, opened 2025-02-19
Delete copy constructor of TfListenerSingleton

furushchev:
Otherwise, the copy constructor returns void which is invalid.

### DIFF
--- a/jsk_recognition_utils/include/jsk_recognition_utils/tf_listener_singleton.h
+++ b/jsk_recognition_utils/include/jsk_recognition_utils/tf_listener_singleton.h
@@ -50,8 +50,8 @@ namespace jsk_recognition_utils
     static tf::TransformListener* instance_;
     static boost::mutex mutex_;
   private:
-    TfListenerSingleton(TfListenerSingleton const&){};
-    TfListenerSingleton& operator=(TfListenerSingleton const&){};
+    TfListenerSingleton(TfListenerSingleton const&) = delete;
+    TfListenerSingleton& operator=(TfListenerSingleton const&) = delete;
   };
 
   // tf Utility


furushchev:
@k-okada 
Regarding changes to support ROS-O, there seems to be a repository `ros-o/jsk_recognition` in addition to `jsk-ros-pkg/jsk_recognition`, but which one (or both?) should I send the PR to?

furushchev:
@k-okada Thank you for the comment.
It might be inaccurate to ask the question here, but how do you use deb packages?
Reading the official docs [1], there are two projects that provide apt repositories:
- https://github.com/v4hn/ros-o-builder
- https://github.com/ubi-agni/ros-builder-action

From what I see, the former tries to use `jsk-ros-pkg/jsk_recognition`, and the latter still uses `ros-o/jsk_recognition`.
I am currently using Ubuntu 24.04, and when I tried to install ROS-O, it seems that the latter is the only option [2], but is it encouraged to migrate the latter project to use `jsk-ros-pkg/jsk_recognition` eventually and dual use both projects?

[1] https://github.com/ros-o/ros-o?tab=readme-ov-file#how-to-run-it
[2] Is that true? Although https://github.com/v4hn/ros-o-builder?tab=readme-ov-file#target-platforms says Ubuntu noble, I couldn't find the apt repository.

k-okada:
@furushchev  Actually, I was not aware of the “ubi-agni” project until @rhaschke  informed me about it last week. (And I apologize for not responding to Robert yet ......) .

I would be very happy if you could send https://github.com/ubi-agni/ros-builder-action pull reqeuest using 
You can use https://github.com/k-okada/ros-o-builder/blob/build-for-jammy/sources.repos e.

To pass jsk related packages on jammy, more paches ( https://github.com/v4hn/ros-deb-builder-action/pulls and https://github.com/s0/git-publish-subdir- action/pulls ) are needed and is still a work in progress for noble. I've forgotten most of it since it was a week ago, but I need to solve the missing packages for kdl and gazebo.

Last but not least information about eusilsp in 24.04, which was accidentally removed from ubuntu but is now available in ubuntu 25.04. We are also working on bringing it back in 24.04.
https://bugs.launchpad.net/ubuntu/+sources/euslisp/+bugs/2096808

PR #2864, opened 2025-04-15
sound_classification: Fix python packages contain incompatible binary files

furushchev:
This PR addresses the build failure issue caused by the inclusion of binaries incompatible with the target environment when building Debian packages from our ROS package. By using the `--no-binary` option, we ensure that these incompatible binaries are excluded during the build process.

I verified the deb build of all packages in this repository on noetic and ROS-O finished without any error.
(I haven't check any function in the package work.)

Hope it will help with releasing the packages in ROS-O...


k-okada:
humm, it is tested at https://github.com/k-okada/ros-builder-action/actions/runs/14321265227 , could you add CI , something like https://github.com/jsk-ros-pkg/jsk_model_tools/pull/262/files#diff-33cc4c925f7fd36575f5e5b61d1e9c942fea5189e2c67d09720d714e19151404 ?

furushchev:
Thank you for the comments!

I just saw https://github.com/ubi-agni/ros-builder-action/pull/42, and looks like you've already found a workaround, great!
I'll close this PR then.

PR #2878, opened 2025-05-06
do not publish when array_size is zero

knorth55:
When input `BoundingBoxArray`'s `boxes` size is zero, topic will be subscribed with the initial value (such as `dimensions.x = 0`).
This PR stop publishing when input size is zero.

### DIFF
--- a/jsk_pcl_ros_utils/src/bounding_box_array_to_bounding_box_nodelet.cpp
+++ b/jsk_pcl_ros_utils/src/bounding_box_array_to_bounding_box_nodelet.cpp
@@ -83,6 +83,7 @@ namespace jsk_pcl_ros_utils
     int array_size = bbox_array_msg->boxes.size();
     if (array_size == 0)
     {
+      NODELET_WARN_THROTTLE(10, "bbox array size is 0, skip publishing");
       return;
     } else if (index_ < 0) {
       return;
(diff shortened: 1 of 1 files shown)


jsk_pcl_ros_utils/src/bounding_box_array_to_bounding_box_nodelet.cpp
```diff
@@ -81,7 +81,10 @@ namespace jsk_pcl_ros_utils
     bbox_msg.header = bbox_array_msg->header;
 
     int array_size = bbox_array_msg->boxes.size();
-    if (index_ < 0) {
+    if (array_size == 0)
+    {
```
mqcmd196, line 85: Adding the `NODELET_WARN_ONCE` message to notify that the subscribed empty bounding box array is better.

mqcmd196 (changes requested):

jsk_pcl_ros_utils/src/bounding_box_array_to_bounding_box_nodelet.cpp
knorth55: I update to add NODELET_WARN

mqcmd196 (approved):
CI green. LGTM

PR #2881, opened 2025-05-23
add resize and timer_rate option in tile_image to reduce cpu load

knorth55:
This PR, I add resize and timer_duration to reduce CPU load of this node.

Current implementation
- publishing gigantic image with 10 hz
  - gigantic image needs CPU load for compressing

This PR
- add option to resize image
- add option to change publish frequency.


knorth55:
@mqcmd196 Thanks! I updated.

